PR #40, opened 2024-12-31
chore: fix security issues

leoparente:


mfiedorowicz (approved):
LGTM, two CodeQL issues seem like false positive as api keys are obfuscated

PR #45, opened 2025-01-08
Fix port in network_discovery docs

IamLunchbox:
`network_discovery` is bound to port 8073, 8072 is reserved for the `device_discovery` api.


leoparente:
Thanks for the contribution!

leoparente (approved):

PR #48, opened 2025-01-08
Clarify discovery method

IamLunchbox:
Currently, network_discovery uses icmp, as can be seen [here](https://github.com/netboxlabs/orb-discovery/blob/develop/network-discovery/policy/runner.go#L71)

### DIFF
--- a/docs/backends/network_discovery.md
+++ b/docs/backends/network_discovery.md
@@ -1,5 +1,5 @@
 # Network Discovery
-The network discovery backend leverages [NMAP](https://nmap.org/) to scan networks and discover IP information.
+The network discovery backend leverages [NMAP](https://nmap.org/) to scan networks with icmp echo requests and discover IP information.
 
 
 ## Configuration


rboucher-me:
@IamLunchbox we are currently working on leveraging additional discovery capabilities of Nmap, including TCP port scanning, that will make this change inaccurate. Is that something that would interest you? Do you have thoughts on how such discovery should be configured and controlled? (for example, only scan a pre-defined set of ports or have the ability to configure what ports to scan?)

IamLunchbox:
I think that would be a very nice addition to improve the IPAM and observability possibilities of netbox! If orb would automatically populate service facts of a  given IP-adress, this knowledge could, for example, be used to audit for documentation and  configuration errors.

In that regards i'd like to add, that it would be nice if more IPAM fields could be populated by orb. For example, I would like to tag and  PTR the scanned IP-adresses as well. But currently, PTR would require a custom script and all tagging / comments would currently apply for ALL findings, since several policy entries are heavily deduplicated.

I suppose at least the following customizations to a scan could be interesting for many users:
- TCP scans (full tcp-handshake) with icmp being optional 
- UDP scans with icmp being optional
- Tweaking of parallelization / speed - some old devices (e.g. ICS, old FWs with small state tables) might not like too fast scans
- Setting of ip ranges as a list, ranges or using the nmap builtin top 10/100/1000

IamLunchbox:
Another thing, but it might not be scoped for orb agent: It would be nice, if ip-addresses, which disappear, would be removed from netbox again.

In dynamic infrastructures ips might appear and disappear again while the ip-adress prefix in netbox only seemingly fills up.

rboucher-me:
> I think that would be a very nice addition to improve the IPAM and observability possibilities of netbox! If orb would automatically populate service facts of a given IP-adress, this knowledge could, for example, be used to audit for documentation and configuration errors.
> 
> In that regards i'd like to add, that it would be nice if more IPAM fields could be populated by orb. For example, I would like to tag and PTR the scanned IP-adresses as well. But currently, PTR would require a custom script and all tagging / comments would currently apply for ALL findings, since several policy entries are heavily deduplicated.
> 
> I suppose at least the following customizations to a scan could be interesting for many users:
> 
> * TCP scans (full tcp-handshake) with icmp being optional
> * UDP scans with icmp being optional
> * Tweaking of parallelization / speed - some old devices (e.g. ICS, old FWs with small state tables) might not like too fast scans
> * Setting of ip ranges as a list, ranges or using the nmap builtin top 10/100/1000

Currently working on these as improvements to network discovery. Do you use Nmap directly? Do you have your "favourite" set of Nmap command options you use?

IamLunchbox:
I would recommend sticking to sane defaults for service discovery:
- `-sT` - scan using full tcp-handshakes, nmap cli default
- `--top-ports 1000` - target the most popular ports nmap knows, nmap cli default
- `--max-retries 0` -  don't retry when probe isn't answered, this orb-agent will run often enough
-  `--host-timeout 3m` -  abort scanning after 5 minutes to not let orb agent get stuck with slow hosts and run forever

But i did not built anything in netbox regarding portscans, I usually use portscans adhoc to find something in my network.

leoparente:
hi @IamLunchbox we added new options support for the network discovery, and now by default it executes the nmap defaults as well (not Ping scan anymore). You can see the changes in the updated docs: https://github.com/netboxlabs/orb-agent/blob/develop/docs/backends/network_discovery.md

Also, if you want to test the changes you can use the docker develop image:` netboxlabs/orb-agent:develop`. And the change will be present in next orb-agent image release.

I will close this PR as it is not required anymore.

IamLunchbox:
Thank you for updating the documenation and the module! I appreciate the implemented options, especially to only run portscans if really necessary :)

PR #66, opened 2025-02-25
fix: selector applies to all block matches

leoparente:
This pull request includes changes to the `processSelector` function in the `gitConfigManager` and updates to the documentation for the Orb Agent's configuration. The key changes are focused on improving the handling of policy paths and clarifying the documentation.

Improvements to `processSelector` function:

* [`agent/configmgr/git.go`](diffhunk://#diff-2e7ff6a74db68db7bf67abb785553044c99821d929cf7e1d742485bc43decfe8L195-R227): Replaced the list of policy paths with a set to ensure uniqueness, and added logic to convert the set back to a list before returning it. This change ensures that duplicate policy paths are not included in the results. Also ensure that every match applies and not only the first one.

Documentation updates:

* [`docs/configs/git.md`](diffhunk://#diff-394366c4c1632186d717eb52066554546fc54195458442ef622f9bbd0bc7a992L45-R56): Updated the example directory structure to reflect a more organized layout and added a new example for a selector that matches all agents. [[1]](diffhunk://#diff-394366c4c1632186d717eb52066554546fc54195458442ef622f9bbd0bc7a992L45-R56) [[2]](diffhunk://#diff-394366c4c1632186d717eb52066554546fc54195458442ef622f9bbd0bc7a992L79-R84)
* Clarified the description of the `selector` section to indicate that an empty selector matches all agents.

### DIFF
--- a/agent/configmgr/git.go
+++ b/agent/configmgr/git.go
@@ -192,7 +192,10 @@ func (gc *gitConfigManager) processSelector(file *object.File, cfg config.Config
 		return nil, err
 	}
 
-	// Check for matching selector
+	// Use a set (map) to store unique policy paths
+	policyPathsSet := make(map[string]struct{})
+
+	// Iterate through all selectors and collect all matching ones
 	for selectorName, entry := range selectors {
 		matches := true
 		for key, value := range entry.Selector {
@@ -204,18 +207,22 @@ func (gc *gitConfigManager) processSelector(file *object.File, cfg config.Config
 
 		if matches {
 			gc.logger.Info("Selector matched", zap.String("selector", selectorName))
-			policyPaths := make([]string, 0)
 			for _, policy := range entry.Policies {
 				if policy.Enabled != nil && !*policy.Enabled {
 					continue
 				}
-				policyPaths = append(policyPaths, policy.Path)
+				policyPathsSet[policy.Path] = struct{}{} // Add path to set (ensures uniqueness)
 			}
-			return policyPaths, nil
 		}
 	}
 
-	return nil, nil
+	// Convert map keys to a slice
+	var policyPaths []string
+	for path := range policyPathsSet {
+		policyPaths = append(policyPaths, path)
+	}
+
+	return policyPaths, nil
 }
 
 func (gc *gitConfigManager) schedule(cfg config.Config, backends map[string]backend.Backend) {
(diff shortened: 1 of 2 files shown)


rboucher-me (approved):
Like it!

agent/configmgr/git.go
```diff
@@ -192,30 +192,39 @@ func (gc *gitConfigManager) processSelector(file *object.File, cfg config.Config
...
-			if cfgValue, exists := cfg.OrbAgent.Labels[key]; !exists || cfgValue != value {
-				matches = false
-				break
+		// If selector is empty, it matches everything
+		if len(entry.Selector) > 0 {
+			for key, value := range entry.Selector {
+				if cfgValue, exists := cfg.OrbAgent.Labels[key]; !exists || cfgValue != value {
+					matches = false
+					break
+				}
```
mfiedorowicz: Range over nil map or size of 0 causes no iterations, hence that if statement seems redundant

```suggestion
		for key, value := range entry.Selector {
			if cfgValue, exists := cfg.OrbAgent.Labels[key]; !exists || cfgValue != value {
				matches = false
				break
			}
```

agent/configmgr/git.go
```diff
@@ -192,30 +192,39 @@ func (gc *gitConfigManager) processSelector(file *object.File, cfg config.Config
...
-
 		if matches {
 			gc.logger.Info("Selector matched", zap.String("selector", selectorName))
-			policyPaths := make([]string, 0)
 			for _, policy := range entry.Policies {
 				if policy.Enabled != nil && !*policy.Enabled {
 					continue
 				}
-				policyPaths = append(policyPaths, policy.Path)
+				policyPathsSet[policy.Path] = struct{}{}
```
mfiedorowicz, line 217: It looks possible to have a selector with two policies pointing to the same path. Shouldn't we at least log a warning (or fail fast) in such scenario?

agent/configmgr/git.go
leoparente: you are right.  🤦🏽 done

agent/configmgr/git.go
leoparente: added warn message! Thanks

mfiedorowicz (approved):

PR #67, opened 2025-02-26
feat: pktvisor opentelemetry setup

leoparente:


agent/backend/otel/utils.go
```diff
@@ -89,11 +90,19 @@ func (d *openTelemetryBackend) request(url string, payload interface{}, method s
...
-			return err
+			return fmt.Errorf("failed to read response body: %w", err)
+		}
+		if err = json.Unmarshal(body, &payload); err == nil {
+			return nil
+		}
+		if yamlErr := yaml.Unmarshal(body, &payload); yamlErr == nil {
+			return nil
 		}
+		return fmt.Errorf("failed to decode response as JSON or YAML: %w", err)
```
mfiedorowicz: Are we not swallowing `yamlErr` here?

leoparente: good point. I will fix it

agent/backend/otel/utils.go
```diff
@@ -89,11 +90,20 @@ func (d *openTelemetryBackend) request(url string, payload interface{}, method s
...
+			return fmt.Errorf("failed to read response body: %w", err)
+		}
+		if err = json.Unmarshal(body, &payload); err == nil {
+			return nil
+		}
+		var yamlErr error
+		if yamlErr = yaml.Unmarshal(body, &payload); yamlErr == nil {
+			return nil
 		}
+		return fmt.Errorf("failed to decode response as JSON: %w. And YAML: %w", err, yamlErr)
```
mfiedorowicz: nit: 
```suggestion
		return fmt.Errorf("failed to decode response as JSON: %w and YAML: %w", err, yamlErr)
```

mfiedorowicz (approved):

PR #88, opened 2025-04-03
chore: add configmgr basic tests

leoparente:
This pull request introduces new test cases for the `configmgr` package, focusing on the `gitConfigManager`, `localConfigManager`, and the `manager.New` function. The changes primarily involve adding comprehensive unit tests to ensure the proper functioning of these components.

### New Tests for `configmgr` Package:

* **Tests for `gitConfigManager`:**
  - Added a test case `TestGitStart` to validate the `Start` method of `gitConfigManager`. This test includes setting up a temporary Git repository, committing files, and verifying the start process without errors.

* **Tests for `localConfigManager`:**
  - Introduced a test case `TestLocalConfigManager` to ensure the `localConfigManager` behaves correctly under different scenarios, including starting with policies, no policies, and missing backend.

* **Tests for `manager.New` Function:**
  - Added a test case `TestManagerNew` to verify that the `New` function returns the correct manager implementation based on the configuration. This includes tests for `LocalManager`, `GitManager`, and defaulting to `LocalManager` when an unknown type is provided.

### DIFF
--- a/agent/configmgr/git_test.go
+++ b/agent/configmgr/git_test.go
@@ -1,0 +1,120 @@
+package configmgr_test
+
+import (
+	"log/slog"
+	"os"
+	"path/filepath"
+	"testing"
+	"time"
+
+	gitv5 "github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/plumbing/object"
+	"github.com/stretchr/testify/mock"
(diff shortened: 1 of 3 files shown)


jajeffries (approved):
These look like a good starting point. I think we could later build on these to handle more edge cases. Eg. when the selector.yaml file is missing etc

PR #154, opened 2025-07-25
Fix docker run typo

my0373:
First PR, so please let me know if I've not followed the process.

Just a readme typo for the ```docker run``` example.

Replaced ```$(PWD)``` with ```${PWD}``` to evaluate the shell variable instead of execute it. 

Otherwise the docker run fails with an error like

```docker: invalid spec: :/opt/orb/: empty section between colons.```


jajeffries (approved):
Looks good to me! Thanks for the contribution

PR #188, opened 2025-09-29
chore: add parameter for image tag

jajeffries:
This pull request updates the GitHub Actions workflow for the Orb Agent's `develop` branch to make the Docker tag configurable when manually triggering the workflow. The main change is allowing users to specify a custom Docker tag, improving flexibility for development and testing.

Workflow improvements:

* Added a new input parameter `docker_tag` to the `workflow_dispatch` event, enabling users to set the Docker image tag when manually running the workflow.
* Updated the Docker build step to use the specified `docker_tag` input instead of the hardcoded `develop` tag, ensuring the built image is tagged as requested.

### DIFF
--- a/.github/workflows/develop.yaml
+++ b/.github/workflows/develop.yaml
@@ -61,6 +61,6 @@ jobs:
             push: true
             cache-from: type=gha
             cache-to: type=gha,mode=max
-            tags: netboxlabs/orb-agent:${{ github.event.inputs.docker_tag }}
+            tags: netboxlabs/orb-agent:${{ github.event.inputs.docker_tag || 'develop' }}
             build-args: |
               GO_VERSION=${{ env.GO_VERSION }}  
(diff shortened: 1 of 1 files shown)


.github/workflows/develop.yaml
```diff
@@ -55,6 +61,6 @@ jobs:
             push: true
             cache-from: type=gha
             cache-to: type=gha,mode=max
-            tags: netboxlabs/orb-agent:develop
+            tags: netboxlabs/orb-agent:${{ github.event.inputs.docker_tag }}
```
mfiedorowicz: On `push` it won't default to `develop`, hence I recommend following change:


```suggestion
            tags: netboxlabs/orb-agent:${{ github.event.inputs.docker_tag || 'develop' }}
```

jajeffries: Thanks, good spot

mfiedorowicz (approved):

PR #193, opened 2025-10-03
refactors the config manager into clearer modules

jajeffries:
Refactors into separate components for
- auth
- MQTT connection
- jwt parsing
- config manager
- extract message handlers into a separate component


agent/configmgr/fleet/auth.go
```diff
@@ -0,0 +1,125 @@
...
+		}
+	}()
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		fleetManager.logger.Error("failed to read response body", "error", err, "status_code", resp.StatusCode)
+		return nil, fmt.Errorf("failed to read response body: %w", err)
+	}
+
+	if resp.StatusCode != 200 {
```
MicahParks:
```suggestion
	if resp.StatusCode != http.StatusOK {
```

agent/configmgr/fleet/auth.go
```diff
@@ -0,0 +1,125 @@
...
+	data.Set("scope", strings.Join(scopes, " "))
+	data.Set("client_id", clientID)
+	data.Set("client_secret", clientSecret)
+	data.Set("audience", "orb")
+
+	fleetManager.logger.Debug("sending token request", "url", tokenURL, "data", data, "client_id", clientID) //, "client_secret", clientSecret)
+
+	req, err := http.NewRequest("POST", tokenURL, bytes.NewBufferString(data.Encode()))
+	if err != nil {
+		fleetManager.logger.Error("failed to create token request", "error", err, "token_url", tokenURL)
```
MicahParks, line 67: Not important, but something to consider long term.

Perhaps it would be best to use log statements with context (separate PR). I think the use case this could serve is adding metrics to log statements or alerting on error log statements. When those metrics/alerts are created from log statements, they would have access to the incoming request's context, which could contain useful information.

```suggestion
		fleetManager.logger.ErrorContext(ctx, "failed to create token request", "error", err, "token_url", tokenURL)
```

MicahParks (approved):

leoparente (approved):

agent/configmgr/fleet/auth.go
jajeffries: Good suggestion. I didn't know about this. I'll look to add it in where I can in a future PR.

PR #210, opened 2025-10-17
feat: add default yaml config file for fleet config manager

leoparente:
This pull request introduces a new default configuration file for the orb agent, updates the Docker image to include this file, and makes minor dependency adjustments in the Go module file. The main focus is on improving configuration management and ensuring the Docker image has the necessary defaults for deployment.

Configuration management improvements:

* Added a new `default_config.yaml` file that provides default configuration values for the orb agent, including config manager settings, backend definitions, and integration with external services using environment variables.
* Updated the Dockerfile to copy the new `default_config.yaml` into the Docker image at `/opt/orb/default_config.yaml`, ensuring the agent has access to default settings at runtime.

Dependency updates:

* Removed the direct dependency on `go.opentelemetry.io/otel/exporters/otlp/otlplog/otlploghttp` and made `google.golang.org/grpc` an indirect dependency in `go.mod` to clean up unused or unnecessary dependencies. [[1]](diffhunk://#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6L23-L26) [[2]](diffhunk://#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R113)

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -29,7 +29,6 @@ const (
 var (
 	cfgFiles []string
 	debug    bool
-	envFile  string
 )
 
 func init() {
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -8,7 +8,6 @@ import (
 	"os/signal"
 	"syscall"
 
-	"github.com/joho/godotenv"
 	"github.com/spf13/cobra"
 	"gopkg.in/yaml.v3"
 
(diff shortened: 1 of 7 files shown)


cmd/main.go
```diff
@@ -65,10 +67,41 @@ func loadConfig(path string, configData *config.Config) error {
 	return nil
 }
 
+func loadEnvFile(path string) (bool, error) {
+	if _, err := os.Stat(path); err != nil {
```
MicahParks: The `--env-file` flag from `docker run` populates environment variables at container runtime. It will not be present in the container's runtime file system.
https://docs.docker.com/reference/cli/docker/container/run/#env

leoparente: thanks, I will remove it

leoparente: done

jajeffries (approved):
Nice, LGTM!

PR #232, opened 2025-11-17
Update git.md - add policy name to selector.yaml example

manrodrigues:
The `selector.yaml` example file is missing the policy name, which results in an error when you try to run it


leoparente (approved):
nice! LGTM

PR #288, opened 2026-03-12
fix: resolve agent hang on shutdown when stop signal sent during active scan

leoparente:
This pull request refactors the OTLP shutdown logic in the `orbAgent` and improves the logging behavior in the OTLP bridge server. The main focus is on simplifying the shutdown process by removing unnecessary context passing and ensuring consistent logging.

OTLP shutdown refactoring:

* Changed the `shutdownOTLP` method in `agent/agent.go` to no longer require a context parameter, instead always using a background context with timeout. This simplifies shutdown calls and avoids potential nil context issues.
* Updated all calls to `shutdownOTLP` in `agent/agent.go` to remove the context argument, ensuring consistency with the new method signature. [[1]](diffhunk://#diff-62b6ad581fe3a3059ae8c85ef0f31dde4092bfdecfa7d6857c470bcacaa8cc8bL126-R126) [[2]](diffhunk://#diff-62b6ad581fe3a3059ae8c85ef0f31dde4092bfdecfa7d6857c470bcacaa8cc8bR272-L279)

Logging improvements:

* Moved the "OTLP bridge server started" log statement in `agent/otlpbridge/server.go` outside the gRPC serving goroutine, so it always logs when the server starts, regardless of serve errors.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -123,7 +123,7 @@ func (a *orbAgent) startBackends(agentCtx context.Context, cfgBackends map[strin
 			a.otlpShutdown = otlpShutdown
 			defer func() {
 				if err != nil {
-					a.shutdownOTLP(agentCtx)
+					a.shutdownOTLP()
 				}
 			}()
 		}
@@ -269,6 +269,7 @@ func (a *orbAgent) Stop(ctx context.Context) {
 			}
 		}
 	}
+	a.shutdownOTLP()
 	if err := a.configManager.Stop(ctx); err != nil {
 		a.logger.Error("error while stopping config manager", slog.Any("error", err))
 	}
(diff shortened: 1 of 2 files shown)


agent/agent.go
```diff
@@ -123,7 +123,7 @@ func (a *orbAgent) startBackends(agentCtx context.Context, cfgBackends map[strin
 			a.otlpShutdown = otlpShutdown
 			defer func() {
```
jajeffries, line 126: There's another issue here. It took a bit of digging so here's cursor's summary

There's a race condition on the otlpShutdown field that this PR doesn't address. The signal handler goroutine is launched before a.Start() in main.go (line 109 vs 127), so if a signal arrives while startBackends is still executing and a backend fails, both the deferred cleanup in startBackends (line 126) and Stop() (line 272) will call shutdownOTLP() concurrently. Since shutdownOTLP uses a non-atomic read-check-nil-call pattern — it reads a.otlpShutdown into a local variable, checks for nil, then sets the field to nil — both goroutines can read the field as non-nil before either writes nil, causing the shutdown function to be invoked twice. A sync.Once (similar to the closeOnce pattern already used in BridgeServer.Stop) would make this idempotent and race-free.

leoparente: great catch Jim, how about now?

jajeffries (approved):
LGTM

PR #296, opened 2026-03-26
feat: add new metadata to OrbAgentInfo

paulstuart:
It originally just contained the version, but now it contains:

Version
CommitHash
BuildBranch
BuiltTime
BuildModified

Note: building has changed in 2 ways:

* It leverages Go's build info to determine the associated git commit (as well as build time and if modified)
* it also unifies the building of the agent itself so that the docker build uses the makefile as well for a consistent build with proper metadata - meta values (build version and branch) are set via linker (no files needed to embed)


jajeffries (approved):
lgtm

PR #315, opened 2026-04-13
fix: OBS-2208 add publish queue to OTLP bridge for MQTT resilience

paulstuart:
Add a bounded message queue to the OTLP bridge so that messages received before or during MQTT disconnection (e.g. credential rotation) are buffered rather than dropped. Messages are automatically drained once the publisher reconnects.

Key changes:
- Move publish logic from individual handlers into a central enqueue method on BridgeServer with a configurable queue limit (default 1000 messages)
- Add drain loop that flushes queued messages when publisher becomes ready
- Add PublisherAdapter interface to decouple from concrete MQTT client
- Add rotation_test.go with zero-data-loss integration tests
- Add server_test.go with queue bounding and startup tests

### DIFF
--- a/agent/otlpbridge/handlers.go
+++ b/agent/otlpbridge/handlers.go
@@ -2,7 +2,6 @@ package otlpbridge
 
 import (
 	"context"
-	"fmt"
 	"log/slog"
 
 	collectorlogs "go.opentelemetry.io/proto/otlp/collector/logs/v1"
@@ -40,11 +39,6 @@ type traceServer struct {
 }
 
 func (s *traceServer) Export(ctx context.Context, req *collectortrace.ExportTraceServiceRequest) (*collectortrace.ExportTraceServiceResponse, error) {
-	pub := s.bridge.GetPublisher()
-	if pub == nil {
-		return nil, fmt.Errorf("publisher not yet initialized")
-	}
-
 	resources := make([]*resourcev1.Resource, 0, len(req.ResourceSpans))
 	for _, rs := range req.ResourceSpans {
 		if rs != nil {
@@ -52,24 +46,11 @@ func (s *traceServer) Export(ctx context.Context, req *collectortrace.ExportTrac
 		}
 	}
 
-	var topic string
-	if isIngestRequest(resources) {
-		topic = s.bridge.GetIngestTopic()
-		if topic == "" {
-			return nil, fmt.Errorf("ingest topic not yet initialized")
-		}
-	} else {
-		topic = s.bridge.GetTelemetryTopic()
-		if topic == "" {
-			return nil, fmt.Errorf("telemetry topic not yet initialized")
-		}
-	}
-
 	payload, err := s.bridge.enc.Marshal(req)
 	if err != nil {
 		return nil, err
 	}
-	if err := pub.Publish(ctx, topic, payload); err != nil {
+	if err := s.bridge.Enqueue(ctx, isIngestRequest(resources), payload); err != nil {
 		return nil, err
 	}
 	return &collectortrace.ExportTraceServiceResponse{}, nil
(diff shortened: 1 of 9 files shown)


paulstuart:
@codex review

paulstuart:
@codex review

agent/otlpbridge/server.go
```diff
@@ -56,39 +104,94 @@ func buildEncoder(name string) (Encoder, error) {
...
+		if msg.isIngest {
+			topic = s.ingestTopic
+		}
+		s.mu.RUnlock()
+
+		if pub != nil && topic != "" {
+			if err := pub.Publish(s.ctx, topic, msg.payload); err == nil {
+				return
+			} else if s.logger != nil {
+				s.logger.Warn("OTLP publish failed, retrying", "error", err, "backoff", backoff)
```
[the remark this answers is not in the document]
paulstuart: added count to retries

agent/otlpbridge/server.go
```diff
@@ -132,9 +235,15 @@ func (s *BridgeServer) Start(ctx context.Context) error {
...
 	s.closeOnce.Do(func() {
+		// Drain in-flight RPCs first so no Export handler enqueues after the
+		// writer goroutine exits. GracefulStop blocks until all active RPCs
+		// complete, then we cancel the writer context to flush remaining items.
 		if s.ingestGRPCServer != nil {
 			s.ingestGRPCServer.GracefulStop()
 		}
+		if s.cancel != nil {
+			s.cancel()
+		}
```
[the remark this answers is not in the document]
paulstuart: Doesn't matter if shutting down anyway

agent/otlpbridge/server.go
```diff
@@ -56,39 +105,98 @@ func buildEncoder(name string) (Encoder, error) {
...
+	}
 }
 
-// GetIngestTopic returns the current topic (for handlers).
+// publishWithRetry publishes a single message, retrying with exponential
+// backoff when no publisher is available or a publish call fails. Retries
+// are bounded to prevent a single poison message from blocking the queue.
+func (s *BridgeServer) publishWithRetry(msg pendingPublish) {
+	backoff := s.initialBackoff
+	for attempt := 0; attempt < maxPublishRetries; attempt++ {
```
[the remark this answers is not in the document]
paulstuart: pushed fix.

agent/otlpbridge/server_test.go
```diff
@@ -113,3 +113,53 @@ func TestBridgeServer_Start_ErrorMessageFormat(t *testing.T) {
...
+
+	cfg := BridgeConfig{
+		ListenAddr:      ":0",
+		Encoding:        "json",
+		MaxPendingQueue: 5,
+	}
+	bridge, err := NewBridgeServer(cfg, nil, logger)
+	require.NoError(t, err)
+	//nolint:errcheck // err doesn't matter
+	defer bridge.Stop(context.Background())
```
leoparente: [nit] better to do something like 
`defer func() { _ = bridge.Stop(context.Background()) }()`
instead of //nolint

paulstuart: I think it's a failure of the linter to complain about a deferred close, but that's just me.
Changed as advised.

agent/otlpbridge/handlers.go
```diff
@@ -124,23 +87,19 @@ type logsServer struct {
...
-			return nil, err
-		}
-	} else {
-		err := s.publishToTelemetryTopic(ctx, req, pub)
-		if err != nil {
-			return nil, err
-		}
+	}
+
+	payload, err := s.bridge.enc.Marshal(req)
```
jajeffries, line 97: If we're calling the bridge to do something, then passing the result of that to the bridge, maybe the first thing (encoding in this case) should live in the bridge too

jajeffries (approved):
Overall looks good - one minor suggestion

agent/otlpbridge/handlers.go
paulstuart: I see the appeal, but as Enqueue is taking different data types we'd need to find a type signature that allowed those types -- and as the msg queue only cares about bytes its not an unnatural separation as it stands today.

I have 2 other PRs on this repo to address msg queue reliability and I would prefer to land what is good enough to move the solution forward and revisit as seen fit.

PR #317, opened 2026-04-15
fix: clear OTLP bridge publisher before MQTT teardown to prevent message drops

paulstuart:
During credential rotation (or agent reset), the MQTT connection was torn down
while the bridge writer goroutine still held a reference to the old publisher.
Publishes into the dead connection exhausted the 10-retry budget and dropped
messages. Clearing the publisher first causes the writer to wait without
consuming retries until the OnReadyHook binds a fresh publisher on reconnect.


paulstuart:
@codex review

jajeffries (approved):

PR #356, opened 2026-05-11
docs(discovery): document VLAN group scope_site behavior

leoparente:
## Summary
- Document on `defaults.vlan.group` (both device-discovery and snmp-discovery README) that the emitted `ipam.vlangroup` is always scoped via `defaults.site`, with an auto-generated slug, so it dedupes across runs.
- Add a `Group scope_site` row to the device-discovery VLAN field table so operators can see the mapping at a glance.

## Why
Backs the upstream behavior change in netboxlabs/orb-discovery#405 — without this note, an operator looking at the docs only sees a bare `VLAN group` field and has no way to predict that the group gets a scope (or that the slug is auto-generated). Opened as **draft** so it can land in lockstep with the orb-discovery PR.

## Test plan
- [x] Docs render correctly on the orb-agent docs site after merge
- [x] Verify the upstream orb-discovery PR (#405) merges first, since the docs describe its behavior

### DIFF
--- a/docs/backends/device_discovery/README.md
+++ b/docs/backends/device_discovery/README.md
@@ -123,7 +123,7 @@ Current supported defaults:
 | ├─ comments | str | VRF comments |
 | ├─ tags | list | VRF tags |
 | vlan       | map  | VLAN-specific defaults        |
-| ├─ group   | str  | VLAN group name. When set, every emitted VLAN is attached to a `ipam.vlangroup` scoped to `defaults.site`: the group's `scope_site` is populated from `defaults.site` (even when that resolves to the `undefined` sentinel), and the slug is auto-generated from the name so the reconciler dedupes the group across runs. |
+| ├─ group   | str  | VLAN group name. When set, every emitted VLAN is attached to a `ipam.vlangroup` scoped to `defaults.site`: the group's `scope_site` is populated from `defaults.site` |
 | ├─ tenant   | str  | VLAN tenant                  |
 | ├─ role   | str  | VLAN role                      |
 | ├─ description | str  | VLAN description          |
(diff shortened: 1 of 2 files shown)


docs/backends/device_discovery/README.md
```diff
@@ -351,4 +351,5 @@ Prefixes are derived from IP addresses discovered on interfaces. The network add
 |-------|--------|-------|
 | VID | `get_vlans()` → VLAN ID | Auto-collected |
 | Name | `get_vlans()` → VLAN name | Auto-collected |
-| Group / Role / Tenant | **Not collected** | Must be set via `defaults.vlan.*` |
\ No newline at end of file
+| Group / Role / Tenant | **Not collected** | Must be set via `defaults.vlan.*` |
+| Group scope_site | Derived from `defaults.site` | Set on the emitted `ipam.vlangroup` whenever `defaults.vlan.group` is configured, so the reconciler can dedupe the group on `(name, slug, scope_site)` across runs. |
```
leoparente:
```suggestion
```

jajeffries (approved):

docs/backends/device_discovery/README.md
```diff
@@ -123,7 +123,7 @@ Current supported defaults:
 | ├─ comments | str | VRF comments |
 | ├─ tags | list | VRF tags |
 | vlan       | map  | VLAN-specific defaults        |
-| ├─ group   | str  | VLAN group                    |
+| ├─ group   | str  | VLAN group name. When set, every emitted VLAN is attached to a `ipam.vlangroup` scoped to `defaults.site`: the group's `scope_site` is populated from `defaults.site` |
```
jajeffries: minor nitpick - should be an "ipam.vlangroup" rather than "a ipam.vlangroup"

samiura (approved):

PR #358, opened 2026-05-12
fix: updates the list of algorithms used to parse JWT

jajeffries:
This change extends the shared parse allowlist with jose.EdDSA and uses it for both topic-related claims (ParseJWTClaims) and exp-based refresh timing (parseJWTExpiry). Behavior is unchanged for existing algorithms; we still decode claims without signature verification, as before (token is obtained from the trusted token endpoint).

Tests add RawJWTWithClaimsAlg and a case that uses an EdDSA header so regressions are caught.

[] go test -race ./agent/configmgr/fleet/...
[] make fix-lint

### DIFF
--- a/agent/configmgr/fleet/auth.go
+++ b/agent/configmgr/fleet/auth.go
@@ -14,7 +14,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/go-jose/go-jose/v4"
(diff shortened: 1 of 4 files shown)


jajeffries:
@codex review

leoparente (approved):

leoparente:
@jajeffries merge develop on it and the scan should pass

PR #364, opened 2026-05-15
fix(fleet): correct policy-count log; split RPC trace from active count (OBS-1727)

leoparente:
## Summary

Fixes [OBS-1727](https://linear.app/netboxlabs/issue/OBS-1727). The fleet-mode `"successfully processed agent policies"` log line reported `count: 1` even when zero policies were attached to the agent. Root cause: it logged `len(rpc)`, which includes payloads with `Action == "sanitize"` that the loop explicitly skips. A single sanitize payload from the control plane was enough to make it look like a policy had been applied.

The fix splits one overloaded log line into two with cleaner semantics:

- **DEBUG `agent_policy RPC handled`** — protocol-shape diagnostic. Reports `applied` (payloads handed to `ManagePolicy`) and `skipped` (sanitize-action payloads), both always emitted. For operators tracing the fleet protocol.
- **INFO `agent managed policies`** — steady-state signal. Reports `count`, read from `PolicyManager.GetPolicyState()` *after* handling the RPC. This is what the reporter wanted: when no policies are attached, `count: 0`, regardless of any sanitize traffic on the wire.

"Managed" matches the codebase's own terminology (`ManagePolicy`, `policyManager`) and accurately describes what `GetPolicyState()` returns — every policy held in the repo, regardless of running/failed state. The "agent" qualifier makes the subject explicit when the log is read across many agents in a fleet, matching this file's existing vocabulary (`handling agent reset`, `handling agent stop`, `agent_policy RPC handled`).

### Behavior

Before (no policies attached, 1 sanitize payload received):
```
INFO  msg="successfully processed agent policies" count=1
```

After (same scenario):
```
DEBUG msg="agent_policy RPC handled"  applied=0 skipped=1
INFO  msg="agent managed policies"    count=0
```

After (one policy attached, no sanitize traffic):
```
DEBUG msg="agent_policy RPC handled"  applied=1 skipped=0
INFO  msg="agent managed policies"    count=1
```

### Notes on the log change

The previous message `"successfully processed agent policies"` and its `count` field are gone. The new INFO message answers the same operator question (how many policies does this agent have?) more accurately. External dashboards or alerts keyed on the old message text will need updating.

## Test plan

- [x] `go test -race ./agent/configmgr/...` (passes)
- [x] `make fix-lint` (0 issues)
- [x] New `TestMessageHandlers_handleAgentPolicies_LogCounters` covers sanitize-only, apply-only, and mixed payloads — asserts DEBUG-level `applied`/`skipped`, INFO-level managed `count` against repo state, absence of the legacy combined log message, and `ManagePolicy` invocation counts
- [x] Pre-existing handleAgentPolicies tests updated to mock `GetPolicyState`
- [x] Existing `TestMessageHandlers_handleAgentPolicies_SkipsSanitizeAction` still passes

### DIFF
--- a/agent/configmgr/fleet/from_rpc.go
+++ b/agent/configmgr/fleet/from_rpc.go
@@ -159,30 +159,39 @@ func (messaging *Messaging) handleAgentPolicies(rpc []messages.AgentPolicyRPCPay
 		}
 	}
 
+	applied := 0
+	skipped := 0
 	for _, payload := range rpc {
-		if payload.Action != "sanitize" {
-			// If the policy data is a string and Format is "yaml" (or empty), try to unmarshal it as YAML
-			// This handles cases where Format="yaml" or where the backend sends YAML without setting Format
-			if dataStr, ok := payload.Data.(string); ok && dataStr != "" && (payload.Format == "yaml" || payload.Format == "") {
-				var structuredData map[string]any
-				if err := yaml.Unmarshal([]byte(dataStr), &structuredData); err != nil {
-					// If unmarshaling fails, log a warning only if Format was explicitly set to yaml
-					if payload.Format == "yaml" {
-						messaging.logger.Warn("failed to unmarshal YAML policy data",
-							"policy_id", payload.ID,
-							"policy_name", payload.Name,
-							"error", err)
-					}
-					// Continue with original string data - let the backend handle it
-				} else {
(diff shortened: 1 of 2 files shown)


leoparente:
@codex review

manrodrigues (approved):

PR #372, opened 2026-05-20
feat(fleet): handle inbox.packages MQTT message to install bundles…

samiura:
## Context

`filesmgr.Manager` was already implemented in orb-agent with fetch, sha256 verification, extraction, atomic symlink-swap, and crash recovery. This PR adds the **MQTT glue** — the missing piece that routes `inbox.packages` messages from the fleet manager into the existing `filesmgr.Manager`.

---

### New: `agent/configmgr/fleet/messages/packages.go`
Defines the `inbox.packages` message types:
- `PackagesCredentialsRPCFunc = "packages_credentials"`
- `BundleSpec` — `{name, version, url, sha256, expires_at}`
- `PackagesCredentialsRPCPayload` — `{bundles: []BundleSpec}`
- `PackagesCredentialsRPC` — full RPC envelope

### Modified: `agent/configmgr/fleet/from_rpc.go`
- Added `filesManager filesmgr.Manager` field to `Messaging` struct
- Added `filesManager` parameter to `NewMessaging` constructor
- Added `case messages.PackagesCredentialsRPCFunc` to `DispatchToHandlers` switch
- Added `handlePackages(ctx, payload)`:
  - Iterates bundles in payload
  - Calls `filesManager.Ensure(ctx, FileSpec{Name, Version, URL, SHA256, Extract: true})` per bundle
  - Non-fatal: a failed bundle is logged and skipped, others continue installing
  - No-ops on empty payload

### Modified: `agent/configmgr/fleet/connection.go`
- Added `filesManager filesmgr.Manager` parameter to `NewMQTTConnection`
- Threads it through to `NewMessaging`

### Modified: `agent/configmgr/fleet.go`
- Added `fm filesmgr.Manager` parameter to `newFleetConfigManager`
- Threads it through to `fleet.NewMQTTConnection`

### Modified: `agent/configmgr/manager.go`
- Added `fm filesmgr.Manager` parameter to `New`
- Passes it to `newFleetConfigManager` for the `fleet` config manager case
- `local` and `git` config managers are unaffected

### Modified: `agent/agent.go`
- Passes the already-constructed `fm` (from `filesmgr.NewManager`) into `configmgr.New`

### New: `agent/configmgr/fleet/packages_test.go`
4 unit tests, no external dependencies required:
- `TestHandlePackages_HappyPath` — single bundle, Ensure called with correct FileSpec
- `TestHandlePackages_EmptyBundles` — empty payload, Ensure never called
- `TestHandlePackages_PartialFailure` — first bundle fails, second still installs (non-fatal)
- `TestDispatchToHandlers_PackagesCredentials` — full dispatch path via JSON-marshalled RPC

---

## Message schema

```json
{
  "schema_version": "2",
  "func": "packages_credentials",
  "payload": {
    "bundles": [
      {
        "name": "nbl_cisco_meraki",
        "version": "0.2.0",
        "url": "https://example.com/bundle.tar.gz?signed-params",
        "sha256": "abc123...",
        "expires_at": 1234567890
      }
    ]
  }
}
```

The `url` is a presigned URL with a short TTL. The `sha256` is of the fat bundle tarball containing the package and all its dependencies pre-installed.

---

## What filesmgr.Ensure does

For each bundle, `Ensure` will:
1. Check if `<root>/<name>/<version>/` already exists and matches sha256 — skip if so
2. Fetch the tarball via `go-getter` using the presigned URL
3. Verify sha256 against the spec
4. Extract into `<root>/<name>/<version>/`
5. Atomically swap the `current` symlink to the new version
6. Persist state to `state.json` for crash recovery

---

## Test results

```
ok   github.com/netboxlabs/orb-agent/agent/configmgr               11.277s
ok   github.com/netboxlabs/orb-agent/agent/configmgr/fleet          3.160s
ok   github.com/netboxlabs/orb-agent/agent/configmgr/fleet/messages  0.357s
```

Note: `agent/secretsmgr` has pre-existing failures requiring a live Docker+Vault — unrelated to this PR.

### DIFF
--- a/agent/configmgr/fleet/from_rpc.go
+++ b/agent/configmgr/fleet/from_rpc.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"log/slog"
 	"os"
+	"time"
 
 	"gopkg.in/yaml.v3"
 
@@ -22,17 +23,27 @@ type Messaging struct {
 	groupManager  *GroupManager
 	resetChan     chan struct{}
 	filesManager  filesmgr.Manager
+	stopCtx       context.Context
+	stopCancel    context.CancelFunc
 }
 
 // NewMessaging creates a new Messaging
 func NewMessaging(logger *slog.Logger, policyManager policymgr.PolicyManager, resetChan chan struct{}, groupManager *GroupManager, filesManager filesmgr.Manager) *Messaging {
+	stopCtx, stopCancel := context.WithCancel(context.Background())
 	return &Messaging{
 		logger:        logger,
 		policyManager: policyManager,
 		groupManager:  groupManager,
 		resetChan:     resetChan,
 		filesManager:  filesManager,
+		stopCtx:       stopCtx,
+		stopCancel:    stopCancel,
 	}
+}
+
+// Stop cancels any in-flight bundle installations.
+func (messaging *Messaging) Stop() {
+	messaging.stopCancel()
 }
 
 // DispatchToHandlers dispatches the message to the appropriate handler
@@ -109,7 +120,7 @@ func (messaging *Messaging) DispatchToHandlers(ctx context.Context, payload []by
 // handlePackages installs each bundle delivered by filesmanager.
 // Failures are non-fatal: a failed bundle is logged and skipped so that
 // other bundles in the same delivery are still installed.
-func (messaging *Messaging) handlePackages(ctx context.Context, payload messages.PackagesCredentialsRPCPayload) {
+func (messaging *Messaging) handlePackages(_ context.Context, payload messages.PackagesCredentialsRPCPayload) {
 	if messaging.filesManager == nil {
 		messaging.logger.Error("filesManager is nil, cannot install bundles")
 		return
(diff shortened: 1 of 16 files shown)


samiura:
@githubcode-claude-review

leoparente:
@codex review

agent/configmgr/fleet/from_rpc.go
```diff
@@ -91,12 +93,53 @@ func (messaging *Messaging) DispatchToHandlers(ctx context.Context, payload []by
...
+func (messaging *Messaging) handlePackages(ctx context.Context, payload messages.PackagesCredentialsRPCPayload) {
+	if len(payload.Bundles) == 0 {
+		messaging.logger.Debug("packages_credentials received with empty bundle list, nothing to do")
+		return
+	}
+	messaging.logger.Info("installing bundles", "count", len(payload.Bundles))
+	for _, bundle := range payload.Bundles {
+		// TODO: check bundle.ExpiresAt before calling Ensure to avoid
+		// unnecessary download attempts with expired presigned URLs.
+		spec := filesmgr.FileSpec{
```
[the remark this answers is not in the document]
samiura: fixed

leoparente:
@codex review

agent/configmgr/fleet/messages/packages.go
```diff
@@ -0,0 +1,25 @@
+package messages
+
+// PackagesCredentialsRPCFunc is the function name for packages credentials RPC calls
+const PackagesCredentialsRPCFunc = "packages_credentials"
+
+// BundleSpec describes a single bundle delivered by filesmanager
+type BundleSpec struct {
```
leoparente, line 11: I think bundlespec should contain mroe information like

```
// FileSpec describes a file the manager should ensure is present on disk.
type FileSpec struct {
	// Name is a logical key, unique within the manager.
	// Examples: "custom_worker_backend", "orb-worker".
	Name string

	// Version is optional. If set, the file is placed under <root>/<name>/<version>/.
	Version string

	// URL is any URL go-getter accepts (HTTP, S3, git, ...).
	URL string

	// SHA256 of the fetched archive or single-file payload. Required. Lowercase hex.
	SHA256 string

	// Extract indicates whether URL points to an archive that should be
	// extracted (true) or a single file that should be placed as-is (false).
	Extract bool

	// TargetPath optionally overrides the default placement path.
	// Not yet supported; Validate() rejects any non-empty value.
	TargetPath string

	// Mode optionally sets the file permission bits for single-file downloads
	// (Extract == false). Zero means default (0o644 — readable data file).
	// Callers that need an executable should set 0o755 explicitly.
	// Ignored when Extract == true (archive extractor preserves per-entry modes).
	Mode os.FileMode
}
```

samiura: Good call on the alignment — done in 5ad95f3. One nuance though: BundleSpec is the wire format (what orb-pro sends over MQTT), while filesmgr.FileSpec is the internal API. Rather than making them identical, I've added the fields that make sense on the wire (Extract, TargetPath, Mode) with omitempty on the optional ones, so orb-pro can control them per bundle. Extract was previously hardcoded to true in handlePackages — it's now driven by the payload, which is more flexible. TargetPath and Mode are passed through so future callers don't need a code change to use them.

leoparente:
@codex review

leoparente (approved):

PR #382, opened 2026-05-27
fix(fleet): unsubscribe the fully-qualified group topic on group removal

paulstuart:
## Problem

`handleAgentGroupRemoval` unsubscribes the **bare group UUID**, but the agent subscribed to the **fully-qualified topic** `orgs/<org>/groups/<id>` (via `groupTopic(orgID, groupID)` in `handleGroupMemberships`). MQTT `UNSUBSCRIBE` requires an exact topic-filter match, so the removal is a **silent no-op** and the broker subscription leaks for the agent's lifetime.

This is the *only* path that even attempts a group-topic unsubscribe — the `dataset_removed` path (`handleDatasetRemoval`) doesn't — so agents effectively never shed group subscriptions except on a full reconnect resync.

## Fix

Thread `orgID` (already in scope at the `DispatchToHandlers` call site) into `handleAgentGroupRemoval` and rebuild the same topic with `groupTopic`. One-line behavioral change; the subscribe path was already correct.

## Tests

Updated the three `handleAgentGroupRemoval` unit tests and the `DispatchToHandlers/group_removed` table case — they previously **encoded the bug** by expecting the bare id; they now assert the fully-qualified `orgs/<org>/groups/<id>` topic and serve as regression guards. Full `agent/configmgr/fleet` suite passes.

## Context

Found while tracing single-run agent unsubscribe paths for the orb-pro OBS-2939 DB cleanup (proposal: per-agent single-run cleanup). Independent of that refactor and safe to land on its own.


leoparente (approved):

jajeffries (approved):
Good catch!

PR #411, opened 2026-06-22
feat(repo): import orb-discovery backends with full history

leoparente:
## PR 1a — Import orb-discovery backends with full history

First step of the orb-discovery → orb-agent monorepo migration (incremental rollout, straight to `develop`). This PR **only imports the code**; CI, the from-source image build, and per-backend releases come in follow-up PRs.

> [!CAUTION]
> **Merge this with "Create a merge commit" — NOT squash.**
> This PR carries the full orb-discovery history (452 commits via `git merge --allow-unrelated-histories`). Squashing collapses it into one commit and **destroys the preserved history**, which is the entire point of this PR. The repo's usual squash-merge convention does **not** apply here.

### What's in it
- Full orb-discovery history merged under `orb-discovery/`, run through `git-filter-repo` to drop `mcp-server/` and rewrite all paths under `orb-discovery/`.
- Go module paths rewritten `…/orb-discovery/<backend>` → `…/orb-agent/orb-discovery/<backend>`; `go mod tidy` per module.
- `go.work` is **git-ignored** (local multi-module dev convenience; the agent image and CI build the agent as a single module — `-mod=mod` conflicts with workspace mode). `make work` regenerates it; `.dockerignore` keeps a local copy out of the image build context.
- 159 per-backend `<backend>/v*` tags carried locally (see "Not in this PR").

### Deliberately NOT in this PR
- **Backend release tags are not pushed yet.** The 159 `<backend>/v*` tags exist locally; pushing them to the remote is a separate step needed by the per-backend release PR.
- Dockerfile stays on published backend images (the from-source switch is a later PR — orb-discovery keeps publishing until then).
- Per-backend lint/test CI (next PR) and per-backend releases (later PR).
- The merged `orb-discovery/.github/` workflows are **inert reference copies** (GitHub only runs root `.github/workflows/`); they're ported later.

### Review surface
The diff is large because it's a history import. The substantive human-authored change is the post-merge mechanics: the module-path rewrite + `go.work` tooling + `.dockerignore` (commits `a571daa`, `e51e526`, `a7a54dd`). The import itself is a faithful filtered mirror of orb-discovery.

### Verified
- Agent module byte-unchanged (`agent/`, `cmd/`, root `go.mod`/`go.sum`).
- Each backend builds + tidies in module mode (`GOWORK=off`); agent `make test` green.
- Agent Dockerfile unchanged; `docker build --target builder` succeeds with the new context.
- 159 backend tags disjoint from the 17 agent `v*` tags (no collision).

PR #412, opened 2026-06-23
ci(repo): per-backend lint and test workflows for discovery backends

leoparente:
## PR 1b — Per-backend lint/test CI

Second step of the orb-discovery monorepo migration. Adds each backend's own lint+test jobs to orb-agent CI, **path-filtered** to `orb-discovery/<backend>/**`, so a backend-only change runs only that backend's jobs (plus the agent's existing `build-and-scan`, which every PR triggers).

> [!NOTE]
> Unlike #411, this is an ordinary PR — **squash-merge is fine** (no history to preserve).

### What's in it
Ported the 8 per-backend lint/test workflows from the imported `orb-discovery/.github/workflows/` (inert reference) to the repo-root `.github/workflows/`, with mechanical path adaptations:
- **Go** (`network`, `snmp`, `gnmi`): `working-directory: orb-discovery/<backend>`, `go-version-file` prefixed, `GOWORK: "off"` (test each module against its own go.mod), golangci `--config ../.github/golangci.yaml` (resolves to `orb-discovery/.github/golangci.yaml`). network-discovery keeps nmap+setcap; snmp/gnmi use `make install-dev-tools`.
- **Python** (`device-discovery`, `worker`): pytest+ruff matrix over Python 3.11–3.14, paths prefixed.
- `actions/checkout`/`setup-go` pinned to the SHAs already used in this repo.

### Verified
- All 8 YAML files parse.
- Go backends: `GOWORK=off go build ./... && make test-coverage` pass for network/snmp/gnmi.
- device-discovery: `pip install .[test]` resolves, pytest collects 1948 tests.
- Fixed a latent bug while porting: network/snmp `find-comment` step now guarded to `pull_request` events (it would otherwise fail on `push`); matches the gnmi + agent `tests.yaml` pattern.

### Not in this PR
- Per-backend releases + PyPI (later PR).
- `orb-discovery/.github/` reference copies are left untouched (ported by the release PR).
- `actions/setup-python` left at `@v6` (repo has no prior pin to reuse; Dependabot can pin it).

### DIFF
--- a/.github/workflows/device-discovery-lint-tests.yaml
+++ b/.github/workflows/device-discovery-lint-tests.yaml
@@ -1,0 +1,62 @@
+name: Device Discovery - lint and tests
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - "orb-discovery/device-discovery/**"
+  push:
+    branches:
+      - "!release"
+    paths:
+      - "orb-discovery/device-discovery/**"
+
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: false
+
+permissions:
+  contents: write
+  pull-requests: write
+
+env:
+  BE_DIR: orb-discovery/device-discovery
+
+jobs:
+  tests:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    strategy:
+      matrix:
+        python: [ "3.11", "3.12", "3.13", "3.14" ]
+    defaults:
+      run:
+        working-directory: ${{ env.BE_DIR }}
+    steps:
+    - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+    - name: Setup Python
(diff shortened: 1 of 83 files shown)


leoparente:
Fixed in 9fc4137 — switched the release-only push exclusion from `branches: ["!release"]` to `branches-ignore: [release]` in all six backend workflows (device/worker lint-tests, network/snmp tests + lint); gnmi already used the correct form. You're right that the `!`-only list never matches on push.

Also rebased the branch onto `develop` now that #413 (from-source image) has landed, so the previously-red `Build & Scan` builds the clean from-source image.

Note: the agent's own `tests.yaml`/`lint.yaml` carry the same pre-existing `branches: ["!release"]` pattern — left out of this PR's scope, worth a separate `ci(repo)` fix.

leoparente:
@codex review

samiura (approved):

PR #430, opened 2026-06-26
feat(agent): add gnmi-discovery backend

leoparente:
## Summary

Adds the missing agent-side **`gnmi_discovery`** backend — the sibling to network/snmp/device discovery — so the agent can launch and manage the in-image `gnmi-discovery` binary, drive its policies, and report its status/version.

The `gnmi-discovery` binary shares network-discovery's exact `/api/v1/{status,capabilities,policies}` REST API and diode flag set, so the new backend mirrors `agent/backend/networkdiscovery` and adds the gNMI-specific surface on top.

## What's included

- **`agent/backend/gnmidiscovery/gnmi_discovery.go`** — implements `backend.Backend` (Configure/Start/Stop/FullReset/Version/GetCapabilities/GetRunningStatus/GetInitialState/ApplyPolicy/RemovePolicy/GetStartTime) + the optional `GetPolicyStatus`. Registers as `gnmi_discovery`, execs `gnmi-discovery`, REST default port **8075** (conflict-free vs 8070/8071/8072/8073/10222/10853).
- **Full gNMI config surface** on top of the network-shared flags (host/port + diode target/client-id/secret/app-name-prefix + dry-run + log-level + otel-endpoint): `profiles_dir` → `--profiles-dir`, `otel_export_period` → `--otel-export-period`, `log_format` → `--log-format`.
- Its own logfmt log normalizer (the established per-backend pattern).
- **`cmd/main.go`** — `gnmidiscovery.Register()` wired in (alphabetical).
- **`docs/backends/gnmi_discovery.md`** — backend + policy documentation.
- Tests: `gnmi_discovery_test.go` (Start asserts every flag *and value* incl. the gNMI extras; otel-target-without-creds; dry-run omits creds but keeps host/port; completed-process error path; the update path DELETEs the **previous** policy name; `GetPolicyStatus`) and `normalize_test.go`.

**Enablement is opt-in** — add `gnmi_discovery:` under `orb.backends:`. `default_config.yaml` is unchanged.

## Verification

- `go build ./...`, `go vet`, `go test -race ./agent/backend/gnmidiscovery/...` all pass; **golangci-lint: 0 issues**.
- A 3-lens adversarial review (Go correctness/concurrency, binary-contract match, test quality) returned **ship**: flag spellings + REST paths + port verified exact against the `gnmi-discovery` binary; interface complete; `fmt.Sprintf("%v")` safe for the int `otel_export_period`; dry-run/OTLP credential handling matches the binary. The two non-blocking findings (update-path DELETE assertion, `GetPolicyStatus` coverage) were addressed by strengthening the tests.

_Note:_ an unsynchronized read/write on the backend's `proc`/`ctx` fields between the monitor goroutine and `Start`/`FullReset` exists, but it is inherited verbatim from the other discovery backends (a shared-pattern issue, not introduced here) — tracked separately to preserve parity.

### DIFF
--- a/agent/backend/gnmidiscovery/gnmi_discovery.go
+++ b/agent/backend/gnmidiscovery/gnmi_discovery.go
@@ -1,0 +1,610 @@
+package gnmidiscovery
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"fmt"
+	"log/slog"
+	"net/http"
+	"sort"
+	"strings"
+	"time"
+
+	"gopkg.in/yaml.v3"
+
+	"github.com/netboxlabs/orb-agent/agent/backend"
+	"github.com/netboxlabs/orb-agent/agent/config"
+	"github.com/netboxlabs/orb-agent/agent/filesmgr"
+	"github.com/netboxlabs/orb-agent/agent/policies"
+	"github.com/netboxlabs/orb-agent/agent/redact"
+)
+
+var _ backend.Backend = (*gnmiDiscoveryBackend)(nil)
+
+const (
+	versionTimeout      = 2
+	capabilitiesTimeout = 5
+	readinessBackoff    = 10
+	applyPolicyTimeout  = 10
+	removePolicyTimeout = 20
+	statusTimeout       = 5
+	defaultExec         = "gnmi-discovery"
+	defaultAPIHost      = "localhost"
+	defaultAPIPort      = "8075"
+)
+
+type gnmiDiscoveryBackend struct {
+	logger     *slog.Logger
+	policyRepo policies.PolicyRepo
+	exec       string
+
+	apiHost     string
+	apiPort     string
+	apiProtocol string
+
+	diodeTarget          string
+	diodeClientID        string
+	diodeClientSecret    string
+	diodeAppNamePrefix   string
+	diodeOtelEndpoint    string
+	diodeTargetFromOtel  bool
+	diodeDryRun          bool
+	diodeDryRunOutputDir string
+	diodeLogLevel        string
+
+	// gNMI-specific options
+	profilesDir      string
+	otelExportPeriod string
+	logFormat        string
+
+	startTime  time.Time
+	proc       backend.Commander
+	statusChan <-chan backend.CmdStatus
+	cancelFunc context.CancelFunc
+	ctx        context.Context
+}
+
+type info struct {
+	Version   string  `json:"version"`
+	UpTimeMin float64 `json:"up_time_seconds"`
+}
+
+// Register registers the gNMI discovery backend
+func Register() bool {
+	backend.Register("gnmi_discovery", &gnmiDiscoveryBackend{
+		apiProtocol: "http",
+		exec:        defaultExec,
+	})
+	return true
+}
+
+func (d *gnmiDiscoveryBackend) Configure(logger *slog.Logger, repo policies.PolicyRepo,
+	config map[string]any, common config.BackendCommons, _ filesmgr.Manager,
+) error {
+	d.logger = logger.With("backend", "gnmi_discovery")
+	d.policyRepo = repo
+	d.diodeTargetFromOtel = false
+
+	var prs bool
+	if d.apiHost, prs = config["host"].(string); !prs {
+		d.apiHost = defaultAPIHost
+	}
+	if port, prs := config["port"]; prs {
+		d.apiPort = fmt.Sprintf("%v", port)
+	} else {
+		d.apiPort = defaultAPIPort
+	}
+
+	d.diodeTarget = common.Diode.Target
+	d.diodeClientID = common.Diode.ClientID
+	d.diodeClientSecret = common.Diode.ClientSecret
+	d.diodeAppNamePrefix = common.Diode.AgentName
+	d.diodeDryRun = common.Diode.DryRun
+	d.diodeDryRunOutputDir = common.Diode.DryRunOutputDir
+
+	if target, prs := config["target"].(string); prs {
+		d.diodeTarget = target
+	}
+	if clientID, prs := config["client_id"].(string); prs {
+		d.diodeClientID = clientID
+	}
+	if clientSecret, prs := config["client_secret"].(string); prs {
+		d.diodeClientSecret = clientSecret
+	}
+	if agentName, prs := config["agent_name"].(string); prs {
+		d.diodeAppNamePrefix = agentName
+	}
+	if dryRun, prs := config["dry_run"].(bool); prs {
+		d.diodeDryRun = dryRun
+	}
+	if dryRunOutputDir, prs := config["dry_run_output_dir"].(string); prs {
+		d.diodeDryRunOutputDir = dryRunOutputDir
+	}
+	if logLevel, prs := config["log_level"].(string); prs {
+		d.diodeLogLevel = logLevel
+	} else if debug, prs := config["debug"].(bool); prs && debug {
+		d.diodeLogLevel = "debug"
+	} else if logger.Enabled(context.Background(), slog.LevelDebug) {
+		d.diodeLogLevel = "debug"
+	}
+
+	// gNMI-specific options
+	if profilesDir, prs := config["profiles_dir"].(string); prs {
+		d.profilesDir = profilesDir
+	}
+	if period, prs := config["otel_export_period"]; prs {
+		d.otelExportPeriod = fmt.Sprintf("%v", period)
+	}
+	if logFormat, prs := config["log_format"].(string); prs {
+		d.logFormat = logFormat
+	}
+
+	if common.Otlp.Grpc != "" {
+		d.diodeOtelEndpoint = common.Otlp.Grpc
+		d.logger.Info("gnmi-discovery using OTLP endpoint",
+			"endpoint", d.diodeOtelEndpoint)
+	}
+	if d.diodeTarget == "" && d.diodeOtelEndpoint != "" {
+		d.diodeTarget = d.diodeOtelEndpoint
+		d.diodeTargetFromOtel = true
+	}
+
+	return nil
+}
+
+func (d *gnmiDiscoveryBackend) Version() (string, error) {
+	var info info
+	url := fmt.Sprintf("%s://%s:%s/api/v1/status", d.apiProtocol, d.apiHost, d.apiPort)
+	err := backend.CommonRequest("gnmi-discovery", d.proc, d.logger, url, &info, http.MethodGet,
+		http.NoBody, "application/json", versionTimeout, "detail")
+	if err != nil {
+		return "", err
+	}
+	return info.Version, nil
+}
+
+func (d *gnmiDiscoveryBackend) Start(ctx context.Context, cancelFunc context.CancelFunc) error {
+	d.startTime = time.Now()
+	d.cancelFunc = cancelFunc
+	d.ctx = ctx
+
+	dOptions := []string{
+		"--diode-app-name-prefix", d.diodeAppNamePrefix,
+		"--host", d.apiHost,
+		"--port", d.apiPort,
+	}
+	if d.diodeDryRun {
+		dOptions = append([]string{
+			"--dry-run",
+			"--dry-run-output-dir", d.diodeDryRunOutputDir,
+		}, dOptions...)
+	} else {
+		opts := []string{
+			"--diode-target", d.diodeTarget,
+		}
+		if !d.diodeTargetFromOtel {
+			opts = append(opts,
+				"--diode-client-id", d.diodeClientID,
+				"--diode-client-secret", d.diodeClientSecret,
+			)
+		}
+		dOptions = append(opts, dOptions...)
+	}
+
+	if d.diodeLogLevel != "" {
+		dOptions = append(dOptions, "--log-level", d.diodeLogLevel)
+		d.logger.Info("gnmi-discovery using log level",
+			"log_level", d.diodeLogLevel)
+	}
+
+	if d.diodeOtelEndpoint != "" {
+		dOptions = append(dOptions, "--otel-endpoint", d.diodeOtelEndpoint)
+		d.logger.Info("gnmi-discovery using OTLP metrics endpoint",
+			"endpoint", d.diodeOtelEndpoint)
+	}
+
+	// gNMI-specific options
+	if d.profilesDir != "" {
+		dOptions = append(dOptions, "--profiles-dir", d.profilesDir)
+		d.logger.Info("gnmi-discovery using profiles dir", "profiles_dir", d.profilesDir)
+	}
+	if d.otelExportPeriod != "" {
+		dOptions = append(dOptions, "--otel-export-period", d.otelExportPeriod)
+	}
+	if d.logFormat != "" {
+		dOptions = append(dOptions, "--log-format", d.logFormat)
+	}
+
+	d.logger.Info("gnmi-discovery startup", "arguments", redact.Args(dOptions))
+
+	d.proc = backend.NewCmdOptions(backend.CmdOptions{
+		Buffered:  false,
+		Streaming: true,
+	}, d.exec, dOptions...)
+	d.statusChan = d.proc.Start()
+
+	// log STDOUT and STDERR lines streaming from Cmd
+	doneChan := make(chan struct{})
+	go func() {
+		defer func() {
+			if doneChan != nil {
+				close(doneChan)
+			}
+		}()
+		stdout := d.proc.GetStdout()
+		stderr := d.proc.GetStderr()
+		for stdout != nil || stderr != nil {
+			select {
+			case line, open := <-stdout:
+				if !open {
+					stdout = nil
+					continue
+				}
+				d.logGnmiDiscoveryOutput(line, slog.LevelInfo)
+			case line, open := <-stderr:
+				if !open {
+					stderr = nil
+					continue
+				}
+				d.logGnmiDiscoveryOutput(line, slog.LevelError)
+			}
+		}
+	}()
+
+	// wait for simple startup errors
+	time.Sleep(time.Second)
+
+	status := d.proc.Status()
+
+	if status.Error != nil {
+		d.logger.Error("gnmi-discovery startup error", "error", status.Error)
+		return status.Error
+	}
+
+	if status.Complete {
+		err := d.proc.Stop()
+		if err != nil {
+			d.logger.Error("proc.Stop error", "error", err)
+		}
+		return errors.New("gnmi-discovery startup error, check log")
+	}
+
+	d.logger.Info("gnmi-discovery process started", "pid", status.PID)
+
+	var version string
+	var readinessErr error
+	for backoff := range readinessBackoff {
+		if status := d.proc.Status(); status.Complete {
+			err := d.proc.Stop()
+			if err != nil {
+				d.logger.Error("proc.Stop error", "error", err)
+			}
+			return errors.New("gnmi-discovery process ended unexpectedly, check log")
+		}
+		version, readinessErr = d.Version()
+		if readinessErr == nil {
+			d.logger.Info("gnmi-discovery readiness ok, got version", "version", version)
+			break
+		}
+		backoffDuration := time.Duration(backoff) * time.Second
+		d.logger.Info("gnmi-discovery is not ready, trying again with backoff",
+			"backoff backoffDuration", backoffDuration.String())
+		time.Sleep(backoffDuration)
+	}
+
+	if readinessErr != nil {
+		d.logger.Error("gnmi-discovery error on readiness", "error", readinessErr)
+		err := d.proc.Stop()
+		if err != nil {
+			d.logger.Error("proc.Stop error", "error", err)
+		}
+		return readinessErr
+	}
+
+	return nil
+}
+
+func (d *gnmiDiscoveryBackend) logGnmiDiscoveryOutput(line string, fallback slog.Level) {
+	trimmed := strings.TrimSpace(line)
+	if trimmed == "" {
+		return
+	}
+
+	msg := trimmed
+	attrs := []slog.Attr(nil)
+	level := fallback
+
+	if parsedMsg, parsedAttrs, parsedLevel, ok := normalizeGnmiDiscoveryLine(trimmed, fallback); ok {
+		msg = parsedMsg
+		attrs = parsedAttrs
+		level = parsedLevel
+	}
+
+	ctx := d.ctx
+	if ctx == nil {
+		ctx = context.Background()
+	}
+
+	d.logger.LogAttrs(ctx, level, msg, attrs...)
+}
+
+func (d *gnmiDiscoveryBackend) Stop(ctx context.Context) error {
+	d.logger.Info("routine call to stop gnmi-discovery", "routine", ctx.Value(config.ContextKey("routine")))
+	defer d.cancelFunc()
+	backend.StopProcess(d.logger, d.proc, d.statusChan, backend.DefaultStopGracePeriod, "gnmi_discovery")
+	return nil
+}
+
+func (d *gnmiDiscoveryBackend) FullReset(ctx context.Context) error {
+	// force a stop, which stops scrape as well. if proc is dead, it no ops.
+	if state, _, _ := backend.GetRunningStatus(d.proc); state == backend.Running {
+		if err := d.Stop(ctx); err != nil {
+			d.logger.Error("failed to stop backend on restart procedure", "error", err)
+			return err
+		}
+	}
+	// for each policy, restart the scraper
+	backendCtx, cancelFunc := context.WithCancel(context.WithValue(ctx, config.ContextKey("routine"), "gnmi-discovery"))
+	// start it
+	if err := d.Start(backendCtx, cancelFunc); err != nil {
+		d.logger.Error("failed to start backend on restart procedure", "error", err)
+		return err
+	}
+	return nil
+}
+
+func (d *gnmiDiscoveryBackend) GetStartTime() time.Time {
+	return d.startTime
+}
+
+func (d *gnmiDiscoveryBackend) GetCapabilities() (map[string]any, error) {
+	caps := make(map[string]any)
+	url := fmt.Sprintf("%s://%s:%s/api/v1/capabilities", d.apiProtocol, d.apiHost, d.apiPort)
+	err := backend.CommonRequest("gnmi-discovery", d.proc, d.logger, url, &caps, http.MethodGet,
+		http.NoBody, "application/json", capabilitiesTimeout, "detail")
+	if err != nil {
+		return nil, err
+	}
+	return caps, nil
+}
+
+func (d *gnmiDiscoveryBackend) GetRunningStatus() (backend.RunningStatus, string, error) {
+	// first check process status
+	runningStatus, errMsg, err := backend.GetRunningStatus(d.proc)
+	// if it's not running, we're done
+	if runningStatus != backend.Running {
+		return runningStatus, errMsg, err
+	}
+	// if it's running, check REST API availability too
+	if _, aiErr := d.Version(); aiErr != nil {
+		// process is running, but REST API is not accessible
+		return backend.BackendError, "process running, REST API unavailable", aiErr
+	}
+	return runningStatus, "", nil
+}
+
+func (d *gnmiDiscoveryBackend) GetInitialState() backend.RunningStatus {
+	return backend.Unknown
+}
+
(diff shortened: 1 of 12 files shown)


cmd/main.go
```diff
@@ -34,6 +35,7 @@ var (
 
 func init() {
 	devicediscovery.Register()
+	gnmidiscovery.Register()
```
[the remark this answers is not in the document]
leoparente: Fixed in 37bafde — good catch, the binary was never bundled. Added a from-source `gnmi-discovery` build stage to `agent/docker/Dockerfile` (mirroring the `network-discovery`/`snmp-discovery` stages, including the `version/BUILD_*.txt` stamp) plus a final `COPY --from=gnmi-discovery /usr/local/bin/gnmi-discovery /usr/local/bin/gnmi-discovery`. I also threaded the version end-to-end so the bundled binary tracks the released backend: `resolve-backend-versions` now emits a `gnmi` output; `develop.yaml`/`build.yaml` pass `GNMI_DISCOVERY_VERSION` from the inline resolve step; and the orchestrated release passes it via a new `gnmi-version` `workflow_call` input on `_agent-release.yaml`, fed by `release.yaml`'s `resolve-versions` job. Verified the binary builds with the exact `go build ... ./cmd/main.go` the stage uses; CI Build & Scan exercises the full multi-arch image.

docs/backends/gnmi_discovery.md
```diff
@@ -0,0 +1,131 @@
...
+orb:
+  backends:
+    common:
+      diode:
+        target: grpc://192.168.0.100:8080/diode
+        client_id: ${DIODE_CLIENT_ID}
+        client_secret: ${DIODE_CLIENT_SECRET}
+        agent_name: agent01
+    gnmi_discovery:
+      host: 192.168.5.11 # default 0.0.0.0
```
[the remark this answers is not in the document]
leoparente: Fixed in 37bafde. The agent backend defaults `host` to `localhost` (`defaultAPIHost`) when it isn't set in config, so the effective default differs from the binary's own `0.0.0.0`. Updated both the YAML example and the parameter table to say `localhost`.

docs/backends/gnmi_discovery.md
```diff
@@ -0,0 +1,131 @@
...
+      log_level: ERROR # default INFO
+      log_format: JSON # default TEXT
+      profiles_dir: /opt/orb/gnmi-profiles # optional: directory of gNMI profile overrides
+      otel_export_period: 30 # optional: seconds between OpenTelemetry metric exports (default 10)
+```
+
+| Parameter | Type | Required | Description |
+|:---------:|:----:|:--------:|:-----------:|
+| host | str | no | REST API host (default `0.0.0.0`) |
+| port | int | no | REST API port (default `8075`) |
```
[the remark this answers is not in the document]
leoparente: Fixed in 37bafde. The agent backend defaults `host` to `localhost` (`defaultAPIHost`) when it isn't set in config, so the effective default differs from the binary's own `0.0.0.0`. Updated both the YAML example and the parameter table to say `localhost`.

agent/backend/gnmidiscovery/gnmi_discovery.go
```diff
@@ -0,0 +1,610 @@
...
+
+	d.logger.LogAttrs(ctx, level, msg, attrs...)
+}
+
+func (d *gnmiDiscoveryBackend) Stop(ctx context.Context) error {
+	d.logger.Info("routine call to stop gnmi-discovery", "routine", ctx.Value(config.ContextKey("routine")))
+	defer d.cancelFunc()
+	backend.StopProcess(d.logger, d.proc, d.statusChan, backend.DefaultStopGracePeriod, "gnmi_discovery")
+	return nil
+}
```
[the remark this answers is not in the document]
leoparente: Fixed in 37bafde. `Stop` now guards the deferred cancel — `if d.cancelFunc != nil { defer d.cancelFunc() }` — so calling `Stop` before a successful `Start` no longer panics. (The same unguarded pattern exists in the sibling discovery backends; fixed here in the new code, and it can be propagated to the others separately to keep parity.)

agent/backend/gnmidiscovery/gnmi_discovery.go
```diff
@@ -0,0 +1,610 @@
...
+		}
+		version, readinessErr = d.Version()
+		if readinessErr == nil {
+			d.logger.Info("gnmi-discovery readiness ok, got version", "version", version)
+			break
+		}
+		backoffDuration := time.Duration(backoff) * time.Second
+		d.logger.Info("gnmi-discovery is not ready, trying again with backoff",
+			"backoff backoffDuration", backoffDuration.String())
+		time.Sleep(backoffDuration)
```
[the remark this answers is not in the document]
leoparente: Fixed in 37bafde. Renamed the readiness log attribute key to `backoff_duration` (it was `backoff backoffDuration`, which contained a space and is awkward to query in structured backends). Inherited from the sibling backends; fixed in the new code here.

agent/backend/gnmidiscovery/gnmi_discovery.go
```diff
@@ -0,0 +1,610 @@
...
+	d.logger.Debug("gnmi-discovery policy remove", "policy_id", data.ID)
+	var resp any
+	name := data.Name
+	// Since we use Name for removing policies not IDs, if there is a change, we need to remove the previous name of the policy
+	if data.PreviousPolicyData != nil && data.PreviousPolicyData.Name != data.Name {
+		name = data.PreviousPolicyData.Name
+	}
+	url := fmt.Sprintf("%s://%s:%s/api/v1/policies/%s", d.apiProtocol, d.apiHost, d.apiPort, name)
+	err := backend.CommonRequest("gnmi-discovery", d.proc, d.logger, url, &resp, http.MethodDelete,
+		http.NoBody, "application/json", removePolicyTimeout, "detail")
```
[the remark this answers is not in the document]
leoparente: Fixed in 37bafde. `RemovePolicy` now path-escapes the policy name (`neturl.PathEscape(name)`) before interpolating it into the DELETE URL, so names containing spaces or `/` resolve to the correct endpoint. Same inherited pattern as the other backends; fixed here in the new code.

leoparente:
@codex review

agent/docker/Dockerfile
```diff
@@ -137,6 +156,8 @@ COPY --from=network-discovery /usr/local/bin/network-discovery /usr/local/bin/ne
 
 COPY --from=snmp-discovery /usr/local/bin/snmp-discovery /usr/local/bin/snmp-discovery
 
+COPY --from=gnmi-discovery /usr/local/bin/gnmi-discovery /usr/local/bin/gnmi-discovery
```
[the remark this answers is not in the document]
leoparente: Fixed in af6ba19. Added `gnmi_discovery:` to `agent/docker/default_config.yaml` alongside the other discovery backends, so the now-bundled binary is actually started in the default Docker/fleet path. Without it, fleet policies for the backend would hit a registered-but-unstarted backend (and removal paths would call into an unconfigured one), exactly as you noted. Grouped it with the discovery backends (network/device/snmp) rather than treating it as opt-in like opentelemetry_infinity, since it's a discovery peer.

agent/backend/gnmidiscovery/gnmi_discovery.go
```diff
@@ -0,0 +1,613 @@
...
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (d *gnmiDiscoveryBackend) GetPolicyStatus() ([]backend.PolicyStatus, error) {
+	var resp backend.StatusResponse
```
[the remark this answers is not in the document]
leoparente: Fixed in af6ba19. `GetPolicyStatus` no longer decodes straight into the shared `backend.StatusResponse` (whose run field is `targets []string`). It now decodes into a gnmi-specific shape that captures each run's singular `target` and normalizes it into `Targets: []string{target}` before returning, so the per-run target survives `convertToRunData` into the policy repo and fleet heartbeats. Added a regression assertion: the status test now returns a run with `target` and asserts the returned `Runs[0].Targets == ["10.0.0.11:6030"]`.

leoparente:
@codex review

agent/backend/gnmidiscovery/gnmi_discovery.go
```diff
@@ -0,0 +1,659 @@
...
+) error {
+	d.logger = logger.With("backend", "gnmi_discovery")
+	d.policyRepo = repo
+	d.diodeTargetFromOtel = false
+
+	var prs bool
+	if d.apiHost, prs = config["host"].(string); !prs {
+		d.apiHost = defaultAPIHost
+	}
+	if port, prs := config["port"]; prs {
```
davidlanouette: NIT: You could make this code a lot shorter and clearer with a simple helper function. 

```golang
func configStringOrDefault(config map[string]any, name string, default string) string {
	if value, ok := config[name]; ok {
		return fmt.Sprintf("%v", value)
	} else {
		return default
	}
}
...
d.apiPort = configStringOrDefault(config, "port", defaultAPIPort)
```

agent/backend/gnmidiscovery/gnmi_discovery.go
```diff
@@ -0,0 +1,659 @@
...
+	url := fmt.Sprintf("%s://%s:%s/api/v1/status", d.apiProtocol, d.apiHost, d.apiPort)
+	err := backend.CommonRequest("gnmi-discovery", d.proc, d.logger, url, &info, http.MethodGet,
+		http.NoBody, "application/json", versionTimeout, "detail")
+	if err != nil {
+		return "", err
+	}
+	return info.Version, nil
+}
+
+func (d *gnmiDiscoveryBackend) Start(ctx context.Context, cancelFunc context.CancelFunc) error {
```
davidlanouette: NIT: This is a very long function.  Consider refactoring it and splitting out some code into separate functions.  Ex: you could move all the code that builds the options into a separate function.

agent/backend/gnmidiscovery/gnmi_discovery.go
```diff
@@ -0,0 +1,659 @@
...
+	d.startTime = time.Now()
+	d.cancelFunc = cancelFunc
+	d.ctx = ctx
+
+	dOptions := []string{
```
davidlanouette: Q: Why `dOptions` _and_ `opts`?  Since they are all put into `dOptions`, do you need `opts`?

agent/backend/gnmidiscovery/gnmi_discovery.go
```diff
@@ -0,0 +1,659 @@
...
+		Buffered:  false,
+		Streaming: true,
+	}, d.exec, dOptions...)
+	d.statusChan = d.proc.Start()
+
+	// log STDOUT and STDERR lines streaming from Cmd
+	doneChan := make(chan struct{})
+	go func() {
+		defer func() {
+			if doneChan != nil {
```
davidlanouette: Will `doneChan` every be nil?

agent/backend/gnmidiscovery/gnmi_discovery.go
```diff
@@ -0,0 +1,659 @@
...
+				}
+				d.logGnmiDiscoveryOutput(line, slog.LevelInfo)
+			case line, open := <-stderr:
+				if !open {
+					stderr = nil
+					continue
+				}
+				d.logGnmiDiscoveryOutput(line, slog.LevelError)
+			}
+		}
```
davidlanouette, line 247: This is confusing - why are you looping?  Why are you closing `stdout` and `stderr`?

I think some comments would be useful here.

agent/backend/gnmidiscovery/gnmi_discovery.go
```diff
@@ -0,0 +1,659 @@
...
+// Register registers the gNMI discovery backend
+func Register() bool {
+	backend.Register("gnmi_discovery", &gnmiDiscoveryBackend{
+		apiProtocol: "http",
+		exec:        defaultExec,
+	})
+	return true
+}
+
+func (d *gnmiDiscoveryBackend) Configure(logger *slog.Logger, repo policies.PolicyRepo,
```
davidlanouette, line 95: Can you add go docs to all the functions?

davidlanouette (commented):
I'll try to finish reviewing later.

agent/backend/gnmidiscovery/gnmi_discovery.go
leoparente: Done in 949e548 — added a `configStringOrDefault(config, name, fallback)` helper (your suggestion) and used it for both `host` and `port`.

agent/backend/gnmidiscovery/gnmi_discovery.go
leoparente: Done in 949e548 — extracted the argument assembly into a `buildArgs()` method, so `Start` is now just build-args → launch → stream → wait-for-ready. (This backend mirrors the existing network/snmp/device backends, which share the same long-Start shape; the same cleanup could be propagated to them in a follow-up.)

agent/backend/gnmidiscovery/gnmi_discovery.go
leoparente: Good catch — `opts` only existed to prepend the diode args ahead of host/port, which was pointless since process flags are order-independent. `buildArgs()` now appends everything to a single slice, so the `opts`/`dOptions` duality is gone.

agent/backend/gnmidiscovery/gnmi_discovery.go
leoparente: No — `doneChan` was a local `make(chan struct{})`, so the `!= nil` guard was always true, and nothing ever received from it: dead code. Removed it in 949e548; the goroutine now just streams the output.

agent/backend/gnmidiscovery/gnmi_discovery.go
leoparente: Added a comment in 949e548. The loop drains the process's stdout and stderr: a closed stream reads as `(_, false)`, so we nil that channel and the `select` then blocks only on the still-open one; the loop exits once both are nil (process gone). This is the standard "select over two channels until both close" pattern.

agent/backend/gnmidiscovery/gnmi_discovery.go
leoparente: Done in 949e548 — added godoc comments to the backend's functions (Configure, Version, Start, Stop, FullReset, GetStartTime, GetCapabilities, GetRunningStatus, GetInitialState, ApplyPolicy, RemovePolicy, GetPolicyStatus, plus buildArgs/configStringOrDefault and the logfmt helpers).

agent/backend/gnmidiscovery/gnmi_discovery.go
```diff
@@ -0,0 +1,690 @@
...
+	config map[string]any, common config.BackendCommons, _ filesmgr.Manager,
+) error {
+	d.logger = logger.With("backend", "gnmi_discovery")
+	d.policyRepo = repo
+	d.diodeTargetFromOtel = false
+
+	d.apiHost = configStringOrDefault(config, "host", defaultAPIHost)
+	d.apiPort = configStringOrDefault(config, "port", defaultAPIPort)
+
+	d.diodeTarget = common.Diode.Target
```
davidlanouette: Can't this also use `configStringOrDefault`?

`d.diodeTarget = configStringOrDefault(config, "target", common.Diode.Target)`

Same with all of the others.

agent/docker/default_config.yaml
```diff
@@ -20,4 +20,5 @@ orb:
     network_discovery:
     device_discovery:
     snmp_discovery:
+    gnmi_discovery:
```
leoparente: ### 
```suggestion
```

agent/backend/gnmidiscovery/gnmi_discovery.go
```diff
@@ -0,0 +1,690 @@
...
+// GetInitialState returns the backend's initial running state (Unknown).
+func (d *gnmiDiscoveryBackend) GetInitialState() backend.RunningStatus {
+	return backend.Unknown
+}
+
+// ApplyPolicy applies a policy via the REST API, removing the prior version
+// first when this is an update.
+func (d *gnmiDiscoveryBackend) ApplyPolicy(data policies.PolicyData, updatePolicy bool) error {
+	if updatePolicy {
+		// To update a policy it's necessary first remove it and then apply a new version
```
davidlanouette, line 391: You can't just do a POST with the correct data?

agent/backend/gnmidiscovery/gnmi_discovery.go
```diff
@@ -0,0 +1,690 @@
...
+		if err := d.RemovePolicy(data); err != nil {
```
davidlanouette, line 392: If `RemovePolicy` fails, do you still want to continue trying to applying it?

davidlanouette, line 392: Should we check if the policy exists before we try to remove it?

agent/backend/gnmidiscovery/gnmi_discovery.go
```diff
@@ -0,0 +1,690 @@
...
+	attrs := []slog.Attr(nil)
+	level := fallback
+
+	if parsedMsg, parsedAttrs, parsedLevel, ok := normalizeGnmiDiscoveryLine(trimmed, fallback); ok {
+		msg = parsedMsg
+		attrs = parsedAttrs
+		level = parsedLevel
+	}
+
+	ctx := d.ctx
```
davidlanouette: consider passing in the context and not storing it on the struct.  (this is the only place it's used, and it's only set in `Start`.  And, this function is only called from `Start`).

agent/backend/gnmidiscovery/gnmi_discovery.go
```diff
@@ -0,0 +1,690 @@
...
+
+	status := d.proc.Status()
+
+	if status.Error != nil {
+		d.logger.Error("gnmi-discovery startup error", "error", status.Error)
+		return status.Error
+	}
+
+	if status.Complete {
+		err := d.proc.Stop()
```
davidlanouette: Should this call `backend.StopProcess(...)` instead?  Or better yet, `d.Stop()`?

agent/backend/gnmidiscovery/gnmi_discovery.go
```diff
@@ -0,0 +1,690 @@
...
+		return errors.New("gnmi-discovery startup error, check log")
+	}
+
+	d.logger.Info("gnmi-discovery process started", "pid", status.PID)
+
+	var version string
+	var readinessErr error
+	for backoff := range readinessBackoff {
+		if status := d.proc.Status(); status.Complete {
+			err := d.proc.Stop()
```
davidlanouette: Should this call `backend.StopProcess(...)` instead?  Or better yet, `d.Stop()`?

agent/backend/gnmidiscovery/gnmi_discovery.go
```diff
@@ -0,0 +1,690 @@
...
+		}
+		backoffDuration := time.Duration(backoff) * time.Second
+		d.logger.Info("gnmi-discovery is not ready, trying again with backoff",
+			"backoff_duration", backoffDuration.String())
+		time.Sleep(backoffDuration)
+	}
+
+	if readinessErr != nil {
+		d.logger.Error("gnmi-discovery error on readiness", "error", readinessErr)
+		err := d.proc.Stop()
```
davidlanouette: Should this call `backend.StopProcess(...)` instead?  Or better yet, `d.Stop()`?

agent/backend/gnmidiscovery/gnmi_discovery.go
```diff
@@ -0,0 +1,690 @@
...
+		apiProtocol: "http",
+		exec:        defaultExec,
+	})
+	return true
+}
+
+// configStringOrDefault returns the stringified value of config[name], or
+// fallback when the key is absent.
+func configStringOrDefault(config map[string]any, name, fallback string) string {
+	if value, ok := config[name]; ok {
```
davidlanouette: a YAML null or wrong type becomes the literal string "<nil>", "12345", or "true" instead of falling back to the default.

https://go.dev/play/p/FZlv_rWSZ_S
vs
https://go.dev/play/p/3zFoerw8hUY

You should check the type of `value` before continuing.

Something like this should work.

```suggestion
	if value, ok := config[name]; ok && value != nil && (reflect.TypeOf(value).Kind() == reflect.String) {
```

agent/backend/gnmidiscovery/gnmi_discovery.go
```diff
@@ -0,0 +1,690 @@
...
+// options. Flags are order-independent, so they are appended in a single pass.
+func (d *gnmiDiscoveryBackend) buildArgs() []string {
+	args := []string{
+		"--diode-app-name-prefix", d.diodeAppNamePrefix,
+		"--host", d.apiHost,
+		"--port", d.apiPort,
+	}
+
+	if d.diodeDryRun {
+		args = append(args, "--dry-run", "--dry-run-output-dir", d.diodeDryRunOutputDir)
```
davidlanouette: Should you check if `d.diodeDryRunOutputDir` is specified before setting it?

agent/backend/gnmidiscovery/gnmi_discovery.go
```diff
@@ -0,0 +1,690 @@
...
+// readQuotedValue reads a quoted logfmt value (handling backslash escapes)
+// starting at the opening quote, returning ok=false if it is unterminated.
+func readQuotedValue(runes []rune, start int) (string, int, bool) {
+	length := len(runes)
+	quote := runes[start]
+	index := start + 1
+	var builder strings.Builder
+
+	for index < length {
+		char := runes[index]
```
davidlanouette, line 637: consider using `if unicode.IsControl(char)` instead.  It's clearer and handles \n, \t, etc.

davidlanouette (commented):
I'm done reviewing.

agent/backend/gnmidiscovery/gnmi_discovery.go
leoparente: There is no get method implemented to check if the policy exists.

agent/backend/gnmidiscovery/gnmi_discovery.go
leoparente: The rest API on the backend does not have "PUT" method, if you POST a policy with a name that already exists, it will fail.

https://github.com/netboxlabs/orb-agent/tree/develop/orb-discovery/gnmi-discovery#routes-v1

agent/backend/gnmidiscovery/gnmi_discovery.go
leoparente: https://github.com/netboxlabs/orb-agent/tree/develop/orb-discovery/gnmi-discovery#routes-v1

agent/backend/gnmidiscovery/gnmi_discovery.go
davidlanouette: ok

agent/backend/gnmidiscovery/gnmi_discovery.go
leoparente: Fixed in 389e921 — `configStringOrDefault` now type-checks: it returns the value only when it is a string (via the comma-ok type assertion `config[name].(string)`), otherwise the fallback. So a YAML null, number, or bool no longer becomes `"<nil>"`/`"12345"`/`"true"`. I used the idiomatic type assertion rather than `reflect` — same effect, no extra import. (port and otel_export_period intentionally keep `fmt.Sprintf("%v")` since they can legitimately be YAML numbers.)

agent/backend/gnmidiscovery/gnmi_discovery.go
leoparente: Done — the string options now use the helper: target/client_id/client_secret/agent_name/dry_run_output_dir each `configStringOrDefault(config, key, common.Diode.X)`, plus profiles_dir/log_format with an empty-string fallback. dry_run (bool) and log_level (special debug fallback) stay explicit, and port/otel_export_period keep Sprintf for numeric YAML values. This is safe now that the helper is string-only (see the L87 thread).

agent/backend/gnmidiscovery/gnmi_discovery.go
leoparente: Yes, intentionally. On an update we remove the old policy and then POST the new one, and the remove is best effort: it can legitimately fail when the policy does not exist yet (for example an "update" that is really a first apply), so aborting there would wrongly block a valid apply. And if the remove fails while the policy still exists, the POST fails too (the backend has no PUT), so ApplyPolicy returns that real error anyway. Nothing is swallowed. This also matches the existing network, snmp, and device discovery backends.

agent/backend/gnmidiscovery/gnmi_discovery.go
leoparente: Done in 73facdd. logGnmiDiscoveryOutput now takes ctx as a parameter, passed from Start's streaming goroutine, and the ctx struct field is removed. It was only set in Start and only read here, so the data flow is now explicit. Behavior is unchanged.

agent/backend/gnmidiscovery/gnmi_discovery.go
leoparente: Done in 73facdd. In dry run we now append --dry-run-output-dir only when a directory is set, so we no longer pass an empty value. (Passing empty previously equaled the binary's own default, so this is mainly for clarity.)

agent/backend/gnmidiscovery/gnmi_discovery.go
leoparente: Switched all three startup and readiness error paths to backend.StopProcess(...), the same graceful helper Stop uses, so cleanup matches a normal stop: the process gets the grace period before being killed, with uniform logging. I kept it as StopProcess rather than d.Stop() because d.Stop() also defers d.cancelFunc() (which would cancel the context Start is still executing inside) and logs "routine call to stop", neither of which fits a startup failure path. Done in 2a6f6a6.

agent/backend/gnmidiscovery/gnmi_discovery.go
leoparente: Switched all three startup and readiness error paths to backend.StopProcess(...), the same graceful helper Stop uses, so cleanup matches a normal stop: the process gets the grace period before being killed, with uniform logging. I kept it as StopProcess rather than d.Stop() because d.Stop() also defers d.cancelFunc() (which would cancel the context Start is still executing inside) and logs "routine call to stop", neither of which fits a startup failure path. Done in 2a6f6a6.

agent/backend/gnmidiscovery/gnmi_discovery.go
leoparente: Switched all three startup and readiness error paths to backend.StopProcess(...), the same graceful helper Stop uses, so cleanup matches a normal stop: the process gets the grace period before being killed, with uniform logging. I kept it as StopProcess rather than d.Stop() because d.Stop() also defers d.cancelFunc() (which would cancel the context Start is still executing inside) and logs "routine call to stop", neither of which fits a startup failure path. Done in 2a6f6a6.

agent/backend/gnmidiscovery/gnmi_discovery.go
leoparente: Good suggestion. This logfmt parser is duplicated from the existing network, snmp, and device discovery backends (it was copied over as part of adding this backend), so I would rather not change only this copy and let it drift from the others. I plan to take a more general approach in a following PR and apply the unicode.IsControl handling across all of them together. The payoff here is also small since these are single line log records, so newlines never appear and a stray tab inside a value is rare, but it is worth doing consistently across the backends.

agent/backend/gnmidiscovery/normalize_test.go
```diff
@@ -0,0 +1,145 @@
...
+			level, ok := parseGnmiDiscoveryLevel(tt.input)
+			assert.Equal(t, tt.ok, ok)
+			if tt.ok {
+				assert.Equal(t, tt.expected, level)
+			}
+		})
+	}
+}
+
+func TestReadUnquotedValue(t *testing.T) {
```
davidlanouette, line 41: Very NIT: You could have all of these `readQuotedValue` tests as a single parameter driven test.

davidlanouette (approved):
Other than one minor NIT, LGTM

agent/backend/gnmidiscovery/normalize_test.go
leoparente: Good nit. Same story as the unicode.IsControl one: this parser and its tests are duplicated across the backends and are moving into a shared backend package in the dedup refactor we are planning. I will write these as a single table driven test there, so it lands once for all backends rather than only in this copy.

It will be my next PR!

PR #439, opened 2026-07-01
feat(agent): report installed bundle state in heartbeats

samiura:
## Summary

Agent heartbeats now report the set of currently installed bundles and their
per-bundle status. Status is intentionally limited to **installed-only** in
this PR — no installing/failed/expired lifecycle states (see Scope below).

## Changes

- **`messages`**: bumps `CurrentHeartbeatSchemaVersion` to `1.2`, adds
  `BundleStateInfo` (state, version, sha256, installed_at) and a new
  `Heartbeat.BundleState` map, keyed by bundle name.
- **`filesmgr.Manager`**: adds `List() []FileEntry` to enumerate all tracked
  entries. Implemented on the base disk engine (backed by the existing
  `store.all()`); `FleetFilesManager` inherits it via embedding. No-op on the
  dummy manager.
- **`configmgr/fleet`**: heartbeater gains a narrow `BundleStateRetriever`
  interface (`List() []filesmgr.FileEntry`), satisfied directly by
  `filesmgr.Manager` — mirrors the existing `GroupRetriever` pattern used for
  group state. `filesManager` (already threaded into `NewMQTTConnection`) is
  now also passed into the heartbeater.

## Design notes

- **No new import cycle.** `configmgr/fleet` already imported `filesmgr` (for
  `NewMessaging`); this just reuses that edge for the heartbeater too.
  `filesmgr` only reaches back to the `messages` leaf package, never to
  `configmgr/fleet`, so there's no cycle risk either direction.
- **Wire fields are deliberately narrow.** `BundleStateInfo` excludes
  `FileEntry.Path` (agent-local FS detail) and `FileEntry.Source` (the raw
  fetch URL — for fleet-delivered bundles this can be a presigned/control-plane
  URL and must not leak into telemetry).
- **`getBundleState` nil-guards `bundleRetriever`** so existing heartbeater
  tests built via struct literal (without every field populated) don't panic;
  production always wires a non-nil retriever (the dummy manager included).

## Scope

This covers installed-only status: a bundle present in the store is reported
as `"installed"` with its version/sha/install time. Lifecycle visibility
(installing / failed / expired) would require recording failed `Ensure`
attempts, which is a materially larger change to the engine — out of scope
here, can be tracked separately if needed.

## Testing

- `make fix-lint && make lint`: 0 issues.
- `make test-all`: all packages touched by this change pass
  (`agent`, `agent/backend/*`, `agent/configmgr`, `agent/configmgr/fleet`,
  `agent/configmgr/fleet/messages`, `agent/filesmgr`).
- `agent/secretsmgr` Vault/CyberArk tests fail locally due to no reachable
  Docker daemon — pre-existing environment gap, unrelated to this change
  (nothing here touches `secretsmgr`).
- Updated `heartbeats_test.go` (new `stubBundleRetriever`, updated
  `newHeartbeater` call sites, schema version assertion) and
  `fleet_messages_test.go` (schema version assertion) for the new signature
  and schema bump. Updated all `filesmgr.Manager` test doubles
  (`mockFilesManager` ×2, `stubDirFilesManager`, `mockEngine`) to satisfy the
  widened interface.

## Follow-ups

- No positive test yet asserting `bundle_state` end-to-end (retriever with
  real entries → marshalled heartbeat). Can add before merge if wanted.
- Confirm the downstream heartbeat consumer accepts schema `1.2` and
  tolerates the additive `bundle_state` field before this ships — the
  schema-version check on receive is currently strict-reject on unknown
  versions.

### DIFF
--- a/agent/agent_test.go
+++ b/agent/agent_test.go
@@ -133,6 +133,8 @@ func (m *mockFilesManager) Ensure(_ context.Context, _ filesmgr.FileSpec) (strin
 func (m *mockFilesManager) Get(_ string) (filesmgr.FileEntry, bool) {
 	return filesmgr.FileEntry{}, false
 }
+
+func (m *mockFilesManager) List() []filesmgr.FileEntry                  { return nil }
 func (m *mockFilesManager) Remove(_ context.Context, _ string) error    { return nil }
 func (m *mockFilesManager) Rollback(_ context.Context, _ string) error  { return nil }
 func (m *mockFilesManager) Subscribe(_ func(filesmgr.FileEvent)) func() { return func() {} }
--- a/agent/backend/worker/worker_filesmgr_test.go
+++ b/agent/backend/worker/worker_filesmgr_test.go
@@ -149,6 +149,8 @@ func (s *stubDirFilesManager) Get(name string) (filesmgr.FileEntry, bool) {
 	}
 	return filesmgr.FileEntry{}, false
 }
+
+func (s *stubDirFilesManager) List() []filesmgr.FileEntry                 { return nil }
 func (s *stubDirFilesManager) Remove(_ context.Context, _ string) error   { return nil }
 func (s *stubDirFilesManager) Rollback(_ context.Context, _ string) error { return nil }
 func (s *stubDirFilesManager) Subscribe(_ func(filesmgr.FileEvent)) func() {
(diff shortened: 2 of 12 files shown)


leoparente:
@codex review

jajeffries (approved):
Generally looks good. Please could we add a pytest integration test validating this? It would be good to be able to see the agent get a bundle and report that back to fleet. Happy for that to be in this PR or in a new one :)

jajeffries (approved):

PR #446, opened 2026-07-09
fix(snmp-discovery): clip interface MTU to NetBox's field limit

### LINKED ISSUE #444: snmp-discovery: interface MTU passed unsanitized
### Bug Report
JUNOS reports MTU `2147483647` on some internal interfaces. NetBox database/API rejects MTU values greater than `65536`, causing these entities to fail validation and failing the entire ingest batch (resulting in persistent failure logs on every ingestion cycle).

### Proposal
The discovery agent should sanitize or cap MTU values to fit NetBox constraints (e.g. ignore or clip values greater than 65536) before sending them to the reconciler.

samiura:
Fixes #444. Some vendors (observed: JUNOS) report 2147483647 (the protocol-level int32 bound) on certain internal interfaces as an 'unlimited' sentinel rather than a real MTU. It passed the existing maxInterfaceMTU check (which guards against corrupt/overflowed SNMP data, not NetBox's schema) and then failed NetBox ingestion downstream, taking out the entire batch on every ingestion cycle.

Adds a separate netboxMaxInterfaceMTU = 65536 cap, enforced just before the value is attached to the entity. Values above it are clipped (with a warning log), not dropped — keeps the interface, corrects only the one field. The existing protocol-level check is left untouched since it protects against a different failure mode.

Updated the existing 'MTU at maximum valid value' test (its input, 2147483647, was already the exact issue #444 scenario) to expect the clipped value, and added a boundary test at exactly 65536 to confirm only values *above* the limit are modified.

### DIFF
--- a/orb-discovery/snmp-discovery/mapping/mappers.go
+++ b/orb-discovery/snmp-discovery/mapping/mappers.go
@@ -30,10 +30,11 @@ const (
 	// exact value on certain internal interfaces as an "unlimited" sentinel rather than a
 	// real MTU, which is why a second, tighter cap (netboxMaxInterfaceMTU) exists below.
 	maxInterfaceMTU = 2147483647
-	// netboxMaxInterfaceMTU is NetBox's actual field constraint (see issue #444). Values
-	// that pass the protocol-level maxInterfaceMTU check above can still exceed this and
-	// fail NetBox ingestion, taking down the entire batch — so this is enforced separately,
-	// just before the value is attached to the entity, rather than folded into the check above.
+	// netboxMaxInterfaceMTU is NetBox's actual field constraint. Values that pass the
+	// protocol-level maxInterfaceMTU check above can still exceed this and fail NetBox
+	// ingestion, taking down the entire batch — so this is enforced separately, just
+	// before the value would be attached to the entity, rather than folded into the
+	// check above.
 	netboxMaxInterfaceMTU = 65536
 )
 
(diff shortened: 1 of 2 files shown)


samiura:
@codex review

leoparente:
@codex review

orb-discovery/snmp-discovery/mapping/mappers.go
```diff
@@ -923,6 +932,16 @@ func (m *InterfaceMapper) Map(values map[ObjectIDIndex]*ObjectIDValue, mappingEn
 							"value", value.Value, "mapping_id", propertyMappingEntry.OID, "interface_index", objectID)
 						continue
 					}
+					// Some vendors (observed: JUNOS) report maxInterfaceMTU itself as an
+					// "unlimited" sentinel on certain internal interfaces. NetBox's actual
+					// field constraint is much lower (issue #444) — clip to it here so one
```
leoparente: dont mention issue in the comment

orb-discovery/snmp-discovery/mapping/mappers.go
```diff
@@ -923,6 +932,16 @@ func (m *InterfaceMapper) Map(values map[ObjectIDIndex]*ObjectIDValue, mappingEn
...
+					// out-of-range interface doesn't fail the entire ingest batch.
+					if mtu > netboxMaxInterfaceMTU {
+						m.logger.Warn("interface MTU exceeds NetBox's field limit, clipping", "mtu", mtu,
+							"netbox_max_mtu", netboxMaxInterfaceMTU, "value", value.Value,
+							"mapping_id", propertyMappingEntry.OID, "interface_index", objectID)
+						mtu = netboxMaxInterfaceMTU
```
leoparente: does we truncate in device discovery or should we ignore the field if it is not in the range?

leoparente: as we are logging the exceed, I think is better to ignore the value, it seems wrong to clip it

leoparente (approved):

PR #451, opened 2026-07-10
fix(device-discovery): harden cisco asa parsing and fxos session prep

leoparente:
## What

Hardens the two Cisco firewall drivers against platform variants that fail discovery today:

1. **`cisco_asa_ssh` — TextFSM parse failures on firmware-variant output.** The strict `cisco_asa` ntc-templates raise `TextFSMError` (State Error) on lines some platforms emit:
   - redundant-interface member lines in `show interface` (`Standby member of Redundant1`, tab-indented on real devices) — the existing pre-filter only covered `Encapsulation:` lines;
   - an extra indented hardware block in `show version` on Firepower appliances running ASA software (`Driver version : 4.12.0` and siblings) — `get_facts` had no pre-filtering at all.

2. **`cisco_fxos` — connect dies at session preparation.** The driver borrows netmiko's `cisco_nxos` device type, whose session prep hardcodes a `terminal width 511` round-trip. FXOS shells authenticate fine but never echo the command, so netmiko raises `ReadTimeout` before any getter runs.

## How

- **Resilient parsing (`cisco_asa_ssh`):** the static pre-filter gains the redundant-member pattern, and a new `_parse_output_resilient` helper wraps all three parse sites (`get_facts`, `get_interfaces`, `get_interfaces_ip`): on `TextFSMError` it extracts the offending line from the exception, strips it (matched on whitespace-stripped content), logs a warning, and retries — bounded at 25 lines, re-raising the original error when the message carries no offending line or nothing was removed. Unknown firmware-variant lines now cost a warn-logged retry instead of failing the whole getter; a line the template State-Errors on could never have parsed anyway.
- **Tolerant session prep (`cisco_fxos`):** a `_TolerantNxosSSH(CiscoNxosSSH)` subclass attempts `terminal width` and paging setup but catches `ReadTimeout` and continues to prompt detection (behavior identical to stock NX-OS when the commands echo; worst case ~30s slower connect on shells that ignore them, then discovery proceeds). `open()` constructs it directly, replicating napalm's `_netmiko_open` semantics (`host=`/`timeout=` kwargs, `_netmiko_device` bookkeeping, `force_no_enable` handling, and the `NetmikoTimeoutException` → `ConnectionException` translation so unreachable devices surface the same error type as every other driver).

## Testing

- New file-based CLI fixtures (documentation-range addressing): `show interface` with tab- and space-indented redundant member lines; `show version` with a multi-line indented hardware block proving sequential strip-and-retry. Both raw fixtures genuinely fail against the real templates without the fix.
- Helper unit tests: strip-once success, re-raise on unhandleable errors, iteration bound.
- FXOS session-prep unit tests: both prep steps timing out → prep completes and prompt detection still runs; clean path unchanged.
- Full `tests/custom_drivers` suite: 1204 passed, 161 skipped — no regressions. `ruff check` clean.


samiura (approved):

samiura:
overall LGTM but some of the tests seems to be failing now.

davidlanouette (approved):

leoparente:
@codex review

PR #452, opened 2026-07-10
fix(device-discovery): pin ntc-templates below 9.2

leoparente:
## What

Pins `ntc-templates` to `~=9.1.0` (>=9.1.0, <9.2.0).

## Why

ntc-templates **9.2.0** (released 2026-07-08) rewrote `hp_procurve_show_system.textfsm` (Filldown values, a new `MEMBER_ID`, restructured states) — `hp_procurve` `get_facts` now fails with a TextFSM State Error. The previous `~=9.1` constraint floats into 9.2.0, so **every fresh CI install of develop fails** `tests/custom_drivers/hp_procurve/test_driver.py::TestProcurveDriver::test_get_facts[normal]` regardless of the PR's content (first observed on #451, reproduced locally: 9.2.0 fails, 9.1.0 green).

9.2.0 also changed `hp_procurve_show_interfaces_brief` and `hp_procurve_show_lldp_info_remote-device_detail` — adapting the driver and fixtures to the new template set should be its own deliberate change, not a hostage of unrelated PRs.


leoparente:
Superseded by #453, which adapts the hp_procurve driver to the 9.2.0 template set and bumps the constraint to ~=9.2.0 — fixing the driver instead of pinning around it.

PR #453, opened 2026-07-10
fix(device-discovery): adapt hp procurve facts parsing to ntc-templates 9.2

leoparente:
## What

Adapts the `hp_procurve` driver to ntc-templates **9.2.0** and moves the constraint to `~=9.2.0`, instead of pinning below it (supersedes #452).

## Why

9.2.0 rewrote `hp_procurve_show_system.textfsm` into a single strict state ending in `^. -> Error`. Real ProCurve devices print a banner line (the model, e.g. `HP ProCurve Switch 2650`) before the "Status and Counters" header; the old template silently skipped that preamble, the new one errors on it — which is why every fresh CI install of develop currently fails `hp_procurve get_facts`. The driver was feeding the template the full raw output while separately harvesting the banner for the model; it now feeds the template only the recognized section (banner handling unchanged).

This is the driver-side adaptation rather than a version pin: the constraint bump in the same change means the tests exercise the actual 9.2.0 template set (which also changed `show_interfaces_brief` and the LLDP detail template — both already pass).

## Testing

- `hp_procurve` suite green under 9.2.0 (was: 1 failure, reproduced deterministically; 9.1.0 masked it via cached wheels locally).
- Full device-discovery suite under 9.2.0: **1798 passed, 161 skipped** — no other template drift.
- The `no-banner` fixture path is unchanged (section scan falls back to line 0).
- `ruff check` clean.


davidlanouette:
Do the older versions still work with this code?

orb-discovery/device-discovery/custom_napalm/hp_procurve.py
```diff
@@ -359,8 +359,20 @@ def get_facts(self) -> dict:
 
         # --- system info via ntc-template ---
         raw_system = self.device.send_command("show system")
+        # The hp_procurve show system template (ntc-templates >= 9.2) is strict
```
davidlanouette, line 362: Are there existing tests for this code that were already failing?

leoparente: yes, The test that is falling in the other PR: 

```
=========================== short test summary info ============================
FAILED tests/custom_drivers/hp_procurve/test_driver.py::TestProcurveDriver::test_get_facts[normal] - textfsm.parser.TextFSMError: State Error raised. Rule Line: 47. Input Line: HP ProCurve Switch 2650
```

https://github.com/netboxlabs/orb-agent/actions/runs/29114038119/job/86432958471?pr=451#step:5:261

orb-discovery/device-discovery/pyproject.toml
```diff
@@ -39,7 +39,7 @@ dependencies = [
     "opentelemetry-sdk~=1.32",
     "opentelemetry-exporter-otlp~=1.32",
     "netmiko~=4.6",
-    "ntc-templates~=9.1",
+    "ntc-templates~=9.2.0",
```
leoparente:
```suggestion
    "ntc-templates~=9.2",
```

samiura (approved):

PR #459, opened 2026-07-15
fix(agent): treat backend 404 on policy removal as already-removed

leoparente:
Closes [OBS-3575](https://linear.app/netboxlabs/issue/OBS-3575/agent-attempts-to-delete-policy-from-backend-even-when-run-is-valid).

## Why

Run-once policies are dropped by the backend after completion, so the agent's cleanup `DELETE` gets a `404 policy not found` and logged a spurious `ERROR` even though the run succeeded and the desired state (policy absent) was reached.

Fixed **agent-side deliberately**: the backend's 404 is truthful REST; it's the agent's cleanup intent ("ensure absent") for which 404 means success. One fix here covers all seven backends — including pktvisor, whose API we don't control — and every backend version already in the field.

## What

- `backend.CommonRequest` now returns a typed `*HTTPError` carrying the status code for all non-2xx responses. Error strings are byte-identical to before — only the type changed, so existing log output and error handling are unaffected.
- `policymgr.RemovePolicy` branches on it: **404 → WARN** ("policy not present on backend at removal; treating as already removed"), anything else keeps the existing ERROR.

Deliberately **WARN, not DEBUG** (ticket offered either): removal deletes by name with rename bookkeeping via `PreviousPolicyData` — a 404 can also mean that bookkeeping broke and a stale policy is still running on the backend under another name. WARN keeps that visible.

Also deliberately **no existence-check-before-delete** (ticket's alternative): TOCTOU race, doubles the API calls, no benefit over tolerating the 404.

## Tests

TDD (watched fail first): 2 tests pinning the typed `HTTPError` from `CommonRequest` via httptest (404 JSON detail + empty-body 500), and 2 policymgr tests asserting 404→WARN-no-ERROR and 500→still-ERROR, both verifying local repo removal proceeds. `go test -race ./agent/...`: 21 packages pass. `make fix-lint` clean.

### DIFF
--- a/agent/backend/utils.go
+++ b/agent/backend/utils.go
@@ -34,6 +34,17 @@ func GetRunningStatus(proc Commander) (RunningStatus, string, error) {
 	return Running, "", nil
 }
 
+// HTTPError is returned by CommonRequest for non-2xx responses so callers can
+// branch on the status code (e.g. treat a 404 on DELETE as already-removed).
+// Message is the fully formatted detail; Error returns it unchanged.
+type HTTPError struct {
+	StatusCode int
+	Message    string
+}
+
+// Error returns the pre-formatted message.
+func (e *HTTPError) Error() string { return e.Message }
+
 // CommonRequest is a generic function to make HTTP requests to the backend
 func CommonRequest(backendName string, proc Commander, logger *slog.Logger, url string, payload any,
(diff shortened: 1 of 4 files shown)


samiura (approved):

leoparente:
@codex review

jajeffries (approved):

PR #460, opened 2026-07-15
fix(snmp-discovery): derive stack member ids with a single scheme per member set

### LINKED ISSUE #458: Bug: Duplicate serial number check drops valid device members
On Cisco Catalyst Switches the ENTITY-MIB::entPhysicalTable has one entry for the switch stack and additional entries for every switch which is part of the stack. The problem is that the switch stack entry uses the same serial number as the master switch of the stack. The duplicate serial number check does drop the chassis and the master switch. Here is a snip of relevant part of a snmpwalk:
iso.3.6.1.2.1.47.1.1.1.1.7.1 = STRING: "c93xx Stack"
iso.3.6.1.2.1.47.1.1.1.1.7.1000 = STRING: "Switch 1"
iso.3.6.1.2.1.47.1.1.1.1.7.2000 = STRING: "Switch 2"
iso.3.6.1.2.1.47.1.1.1.1.4.1 = INTEGER: 0
iso.3.6.1.2.1.47.1.1.1.1.4.1000 = INTEGER: 1
iso.3.6.1.2.1.47.1.1.1.1.4.2000 = INTEGER: 1
iso.3.6.1.2.1.47.1.1.1.1.11.1 = STRING: "FVHxxxxxxxx"
iso.3.6.1.2.1.47.1.1.1.1.11.1000 = STRING: "FVHxxxxxxxx"
iso.3.6.1.2.1.47.1.1.1.1.11.2000 = STRING: "FVHyyyyyyyy"
Log Entry:
{"time":"2026-07-15T12:16:04.817383801Z","level":"WARN","msg":"chassis row dropped: ambiguous duplicate member id","backend":"snmp_discovery","count":"2","id":"1"}
This does not only happen for C9300 switches, but also for other models.
I think this could be solved by ignoring this check if one of the device members detected as duplicate has no parent (root member) and the other one has the root member as parent.

leoparente:
Fixes #458.

## Root cause

Not the duplicate serial: the reporter's stack container row is class 11 and is correctly never a member candidate, so its shared serial is harmless (as are the class-1 StackPort rows sharing stack-cable serials in the full walk). Their attached `entPhysicalTable` shows the members carry **zero-based** `entPhysicalParentRelPos` (0,1,2,3) while `entPhysicalName` carries the true 1-based member numbers. `deriveMemberID` mixed tiers per member — the prel=0 member fell to the name tier (id 1) while its siblings used the prel tier (ids 1,2,3) — so two members collided on id 1 and the ambiguous-id dedup dropped both (including the master), while the survivors kept off-by-one ids that would corrupt `vc_position`.

## Fix

`assignMemberIDs` derives ids with one scheme for the whole member set:

- `parentRelPos` only when every value is positive and distinct;
- else name trailing ints when every member has one and they're distinct (0 allowed — FPC-style members are genuinely zero-numbered);
- **ambiguity is still refused, never renumbered**: duplicate positive positions (or duplicate full-coverage name numbers) with no rescuing signal keep their colliding ids so the existing pass-2 dedup drops those rows exactly as before;
- else ordinal in walk order.

Set-wide tier rejection of a present signal is logged (Info when names rescue cleanly — the corrected path for this device class — Warn for lossy outcomes), with structured fields.

Validated against the reporter's full attached walk: members now get ids 1–4 with nothing dropped. The regression fixture uses non-sequential member numbers (1,2,4,5) so a broken ordinal assignment cannot masquerade as the name tier.

Known accepted limitation (documented in the code): scheme selection runs over pre-dedup rows, so a hypothetical duplicate-serial class-3 row with out-of-scheme prel could downgrade the tier for the set — no real walk exhibiting this is known, and pass-1 serial dedup bounds the damage.

## Corpus scan (LibreNMS, 1,880 real-device recordings)

Every recording with ≥2 chassis-class rows was classified by `parentRelPos` scheme:

- **22 devices with clean 1-based sets** (Cisco SVL pairs, Cisco SB, Eltex, Huawei CE, …) — byte-identical behavior, confirming the no-change guarantee.
- **3 zero-based devices**: Catalyst **9200L** (`{0,1,2}`, "Switch 1..3" — reproduces #458 verbatim today), Catalyst **2960-X** (`{0,2,3}` — the *same box* mixes offset and non-offset prel semantics, proving per-member trust of prel is unfixable in principle), ZTE ZXA10 C320 (no names → ordinal; positions shift `{1,3}`→`{1,2}`, an accepted delta on a device that isn't a switch stack).
- **0 devices with ambiguous duplicate positives** — the preserved refusal path is a rare fail-safe, as expected.
- One accepted delta on a Rittal LCP cooling unit with garbage prel (`{-1,1,1,2,3}`): old code tier-mixed into dropping 3 of 5 rows; new code emits ordinal ids with warns. Neither treatment is meaningful for a non-stack device.

## Lab dry-run validation (orb-test-lab snmpsim, real recordings, real binary)

Ran the branch binary in `--dry-run` via the policy API against LibreNMS recordings replayed by the lab's snmpsim (c9200l serials grafted — LibreNMS anonymizes them to empty, which trips the pre-existing empty-serial drop; container shares member 1's serial exactly like the reporter's device):

| Catalyst 9200L (zero-based `{0,1,2}`) | old (develop) | this PR |
|---|---|---|
| log | `WARN chassis row dropped: ambiguous duplicate member id id=1 count=2` | `INFO member id: … using entPhysicalName-derived ids` |
| output | **1 device carrying Switch 3's serial** — master + Switch 2 dropped, no VC | VirtualChassis + master (member 1's serial) + members at vc_position 2, 3 |

The old failure is worse than the issue describes: the stack reaches NetBox impersonated by its third member. Catalyst 2960-X (unmodified recording, 152 entities) ingests cleanly as a no-regression check on a messier real walk.

## Review & tests

Plan and commit each went through multi-round adversarial review (three plan rounds, two commit rounds). TDD throughout: regression fixture reproducing the reporter's exact walk shape (watched fail: 2 survivors with wrong ids + dropped id 1), TranslateAsStack end-to-end (vc_position/naming), and 13 set-level unit tests covering every tier, both ambiguity-refusal paths, and log contracts. Sentinel tests pinning existing dedup/refusal semantics pass unmodified. Full `go test -race ./...` green; `make fix-lint` clean.

### DIFF
--- a/orb-discovery/snmp-discovery/mapping/chassis.go
+++ b/orb-discovery/snmp-discovery/mapping/chassis.go
@@ -41,7 +41,7 @@ func (m *ChassisInventoryMapper) Map(
 
 // ChassisMember is one row of ENTITY-MIB entPhysicalTable identified
 // as a top-level chassis (class=3, containedIn=0). The ID is the
-// derived logical member id (see deriveMemberID); EntPhysicalIndex is
+// derived logical member id (see assignMemberIDs); EntPhysicalIndex is
 // the raw row index used for entAliasMappingTable chain walks.
 // AssetTag is the trimmed entPhysicalAssetID value ("" when unset or not walked).
 type ChassisMember struct {
@@ -108,9 +108,9 @@ func isStackContainerParent(oids ObjectIDValueMap, idx string) bool {
 //     container; used by Cisco StackWise Virtual on the 9400/9500/9600
 //     series and similar pair architectures).
 //
-// Returns members sorted ascending by ID. Member id derivation uses
-// the full 3-tier precedence in deriveMemberID (ParentRelPos > 0 →
-// trailing-int parse of EntName → ordinal fallback).
+// Returns members sorted ascending by ID. Member ids are derived
+// set-wide by assignMemberIDs (one scheme for all members:
+// parentRelPos → entPhysicalName trailing int → ordinal).
 func extractInventory(oids ObjectIDValueMap, logger *slog.Logger) ChassisInventory {
 	candidates := []string{}
 	for oid, v := range oids {
@@ -147,17 +147,16 @@ func extractInventory(oids ObjectIDValueMap, logger *slog.Logger) ChassisInvento
 		})
 	}
 
-	// Sort by entPhysicalIndex first so the ordinal fallback is
-	// deterministic when neither parentRelPos nor entPhysicalName
-	// provides an id signal.
+	// Sort by entPhysicalIndex first so assignMemberIDs' set-wide
+	// ordinal tier (used when neither the parentRelPos column nor
+	// entPhysicalName is usable for the whole set) assigns
+	// deterministic walk-order ids.
 	slices.SortFunc(members, func(a, b ChassisMember) int {
 		ai, _ := strconv.Atoi(a.EntPhysicalIndex)
 		bi, _ := strconv.Atoi(b.EntPhysicalIndex)
 		return ai - bi
 	})
-	for i := range members {
-		members[i].ID = deriveMemberID(members[i], i+1)
-	}
+	assignMemberIDs(members, logger)
 	// Dedup pass 1: drop later-occurring duplicates of the same serial,
 	// keep the lowest-id occurrence. Track dropped ids and their
 	// entPhysicalIndexes for the routing warn-and-skip rule.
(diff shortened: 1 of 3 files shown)


samiura (approved):

leoparente:
@codex review

PR #461, opened 2026-07-15
feat(snmp-discovery): support device tenant in policy defaults

leoparente:
Fixes [ENGHLP-1479](https://linear.app/netboxlabs/issue/ENGHLP-1479/snmp-discovery-support-device-tenant-in-policy-defaults).

## Why

snmp-discovery had no way to set a tenant on discovered devices: every scan emitted tenant-less devices, which can never match a tenant-owned NetBox device, so each scan created a duplicate with tenant `None`. (Per-entity tenants for `ip_address`/`prefix`/`vlan` already existed — devices were the only entity without the knob.)

## What

Top-level `defaults.tenant`, mirroring device-discovery's placement and semantics:

```yaml
defaults:
  tenant: customer-a            # scalar: tenant name
  # or:
  tenant:
    name: customer-a
    group: customers            # optional: group/description/comments/tags
```

- Applied to **device entities only** — deliberately no cascade to IP/prefix/VLAN tenants (matching device-discovery, where `defaults.tenant` is device-only; the per-entity knobs remain authoritative). Pinned by negative tests.
- `override_defaults.tenant` merges **field-wise** (like `vrf`): a per-target override can refine one field (e.g. `name`) without restating the rest.
- Virtual-chassis members, the master ref, and nested device stubs inherit the tenant through the existing `master.Tenant` copies — no chassis changes needed.

## Upgrade behavior (matcher implications)

- Tenant-owned NetBox devices now match via the diode plugin's `name+site+tenant` rung — the ticket's fix.
- Devices previously created tenant-less by earlier scans are matched via the higher-precedence rungs (`primary_ip4/6`, `asset_tag`, or `virtual_chassis+vc_position` for VC members) and are **updated in place** with the new tenant — no duplicates.
- Caution: a device already owned by a **different** tenant will likewise match via those higher rungs and be silently re-tenanted — operators with mixed-tenant sites should scope tenants via per-target `override_defaults`.

## Review & tests

Plan and commits went through multi-round adversarial review. TDD: config-layer tests (scalar/mapping/null/bad-kind decode, receiver reset, field-wise merge incl. group-only override), device-mapper emission tests (full fields / name-only / unset-stays-nil), no-cascade negative tests for IPAddress and VLAN mappers, and a VC propagation pin on the existing TranslateAsStack test. YAML-alias decoding and per-target `MergeDefaults` copy semantics verified empirically. Full `go test -race ./...` green; `go vet` clean; `make fix-lint` clean.

### DIFF
--- a/docs/backends/snmp_discovery/README.md
+++ b/docs/backends/snmp_discovery/README.md
@@ -90,6 +90,7 @@ SNMP discovery policies are broken down into two subsections: `config` and `scop
 | location | string | no | Default location for discovered devices. Accepts a literal name or an SNMP OID reference (see [Default values from SNMP OIDs](#default-values-from-snmp-oids)) |
 | asset_tag | string | no | Default asset tag for discovered devices. Accepts a literal value or an SNMP OID reference (see [Default values from SNMP OIDs](#default-values-from-snmp-oids)). NetBox enforces a 50-character limit; resolved values longer than 50 characters are warn-logged and skipped |
 | role | string | no | Default role for discovered devices |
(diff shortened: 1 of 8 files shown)


leoparente:
@codex review

samiura (approved):

PR #462, opened 2026-07-17
fix(agent): honor orb.debug.enable from the config file

leoparente:
Fixes [OBS-3432](https://linear.app/netboxlabs/issue/OBS-3432/agent-debug-mode-configured-via-yaml-does-not-enable-debug-logs-only).

## Root cause

`orb.debug.enable` was parsed into the config struct but had **zero consumers**: the logger level was fixed from the `-d` flag before `config.Load` ran, and `agent.New` received the raw flag — so YAML-configured debug produced no debug logs and no backend debug propagation.

## Fix

The logger now sits on a `slog.LevelVar` that is raised to Debug immediately after the config file is parsed (with a debug-level announcement line), and `agent.New` receives `flag OR config` so `BackendCommons.Debug` (worker's `--debug`) and the `logger.Enabled(Debug)` inference in the gnmi/network/snmp backends behave identically to `-d`.

Two behavior notes:

- **Parse-time caveat**: debug-level diagnostics emitted while the config file is still being parsed (the ORB_ overlay's invalid-name/empty-value skip lines) are suppressed when debug comes only from YAML — the level is still Info at that point. `-d` remains the way to capture those. Documented in `docs/config_samples.md`.
- **OTLP deployments**: when the OTLP log bridge is configured, `logger.Enabled(Debug)` already returns true today via the otelslog multi-handler (pre-existing), so the `logger.Enabled` propagation leg only changes behavior in non-OTLP setups.

Side benefit: `ORB_DEBUG__ENABLE=true` (the standard env overlay) now also enables debug, consistent with every other config key.

## Verification

TDD (RED watched: undefined helpers) with 6 unit tests pinning flag-immediate, default-info, yaml-only flip + announcement, no-duplicate-announce, and both-off. **Live-verified end to end** on a built binary: `enable: true` without `-d` emits DEBUG lines (announcement + organic agent debug lines); `enable: false` without `-d` emits zero; `-d` alone is byte-identical to before including the pre-parse ORB_ overlay debug line. `go test -race ./cmd/... ./agent/...` green; `make fix-lint` clean.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"fmt"
+	"io"
 	"log/slog"
 	"os"
 	"os/signal"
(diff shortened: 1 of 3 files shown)


leoparente:
@codex review

manrodrigues (approved):

PR #465, opened 2026-07-20
fix(agent): fail startup on an unsupported secrets_manager type

leoparente:
Fixes [OBS-3431](https://linear.app/netboxlabs/issue/OBS-3431/agent-should-fail-to-start-when-an-unsupported-secrets-manager-type-is).

## Root cause

`secretsmgr.New`'s `default` switch arm returned a no-op `dummyManager` for **both** an empty (unconfigured) `secrets_manager.active` and any unrecognized value. So `ORB_SECRETS_MANAGER__ACTIVE=invalid` (or any typo) started the agent normally while silently resolving no credentials, with no signal the config was invalid — both cases logged the same `no secrets manager specified or invalid type, skipping`.

## Fix

`New` now returns `(Manager, error)` and splits the two cases:
- **empty/unset active** → no-op `dummyManager`, no error, logs `no secrets manager specified, skipping` (unchanged startup behavior);
- **non-empty unsupported value** → error listing the supported types, propagated by `agent.New` so the agent exits at startup (`cmd/main.go` already logs + `os.Exit(1)` on an `agent.New` error).

## Verification

TDD: the existing `TestNew` table's `invalid → dummyManager` case (which encoded the bug) is changed to `invalid → error`; the empty → dummyManager case stays green as the pin. **Live-verified** on a built binary against the ticket's repro: `ORB_SECRETS_MANAGER__ACTIVE=invalid` exits **non-zero** with `unsupported secrets manager type "invalid" (supported: vault, fleet, delinea, doppler, cyberark)`; no env var still logs the skip line and proceeds past secrets init; a set-but-empty value is dropped by the env overlay and also skips. `go test -race ./cmd/... ./agent/...` green (22 pkgs); `make fix-lint` clean.

One small consequence worth noting: a mis-cased or whitespaced value (`Vault`, ` vault `) that previously skipped silently now fails startup. That is desirable, and the docs have only ever specified lowercase type names.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -98,7 +98,11 @@ var _ Agent = (*orbAgent)(nil)
 
 // New creates a new agent
 func New(logger *slog.Logger, c config.Config, debug bool) (Agent, error) {
-	sm := secretsmgr.New(logger, c.OrbAgent.SecretsManager)
+	sm, err := secretsmgr.New(logger, c.OrbAgent.SecretsManager)
+	if err != nil {
+		logger.Error("error during create secrets manager, exiting", "error", err)
+		return nil, err
+	}
 	fm := filesmgr.New(logger, c.OrbAgent.FilesManager)
 	pm, err := policymgr.New(logger, sm, c)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


leoparente:
@codex review

samiura (approved):

PR #470, opened 2026-07-21
feat(agent): report failed bundle installs in heartbeat bundle_state

samiura:
## Title

`Report failed bundle installs in agent heartbeat bundle_state`

### Summary

Today, `bundle_state` in the agent heartbeat only ever reports bundles that
have successfully installed. A failed install (checksum mismatch, the
10-minute install timeout, or a download/extract error) is silently omitted
instead of being reported — consumers have no way to distinguish "still
installing" from "failed and never going to succeed without intervention."

This PR closes that gap on the **agent side**. `filesmgr.Ensure` now records
a per-name failure (in-memory only) whenever an install attempt fails, and
clears it as soon as a later attempt for the same name succeeds. The
heartbeater merges this into `bundle_state`, reporting `state: "failed"` with
an `error` message — following the same `State` + `Error` pattern already
used by `PolicyStateInfo` and `BackendStateInfo`.

### What changed

- **`agent/filesmgr`**: `Ensure` now wraps the existing install logic
  (renamed to `ensureLocked`) to record a `FailureEntry` (name, version,
  error, timeout flag, timestamp) on failure via a new in-memory
  `recordFailure`/`clearFailure`/`ListFailures()` on the `filesmgr` struct,
  exposed through a new `ListFailures() []FailureEntry` method on the
  `Manager` interface (implemented by `dummyFilesManager` as well).
  `FailureEntry` is intentionally **not** persisted to `state.json` — a
  failed attempt carries no on-disk artifact worth surviving a restart, and
  every restart/reconnect already triggers a fresh install attempt via
  `SendBundleListRequest`.
- **`agent/configmgr/fleet/heartbeats.go`**: `BundleStateRetriever` gains
  `ListFailures()`; `getBundleState()` merges failures into the existing
  installed-bundle map. A bundle with a prior successful install but a newer
  failed re-fetch attempt reports `state: failed` with `version`/`error`
  updated to the failed attempt while `sha256`/`installed_at` continue to
  reflect the last successful install (mirrors how `BackendStateInfo` keeps
  `LastRestartTS`/`LastRestartReason` alongside a current error).
- **`agent/configmgr/fleet/messages`**: `CurrentHeartbeatSchemaVersion` bumped
  `1.2` → `1.3`. `BundleStateInfo` gains `Error` (`json:"error,omitempty"`)
  and `FailedAt` (`json:"failed_at,omitzero"`) fields; new
  `BundleStateFailed = "failed"` constant alongside the existing
  `BundleStateInstalled`.
- Test mocks across `agent/agent_test.go`,
  `agent/backend/worker/worker_filesmgr_test.go`,
  `agent/filesmgr/fleet_test.go`, and `agent/configmgr/fleet/packages_test.go`
  updated to satisfy the extended `Manager` interface.
- New coverage: checksum-mismatch and install-timeout failure recording/
  clearing in `agent/filesmgr/filesmgr_test.go`; failed + last-known-good
  merge behavior in `agent/configmgr/fleet/heartbeats_test.go`.

### Design decisions

- Plain error string, not a structured error code/category.
- Failure state clears on the next successful install attempt rather than
  persisting as visible history — matches how `BackendErr`/
  `PolicyStateInfo.Error` already behave.

### Out of scope / follow-up needed

This PR is the **agent-side half only**. The control-plane side — ingesting
the updated `bundle_state` shape and surfacing per-bundle failure status
through whatever API/store consumers read (desired-state / assignment
status) — lives in a separate service and still needs to be scoped there.

### Testing

- `go build ./...` — clean
- `go vet ./...` — clean
- `go test ./agent/filesmgr/... ./agent/configmgr/fleet/...` — pass
- Full `go test ./...` passes aside from pre-existing `agent/secretsmgr`
  Vault/Delinea integration tests, which require a local Docker daemon /
  external services and are unrelated to this change.
</content>

### DIFF
--- a/agent/configmgr/fleet/heartbeats.go
+++ b/agent/configmgr/fleet/heartbeats.go
@@ -21,13 +21,14 @@ const (
 // heartbeatTickInterval is the delay between periodic heartbeats. Tests shorten it.
 var heartbeatTickInterval = heartbeatFreq
 
-// BundleStateRetriever provides read-only access to installed and failed
-// bundle state for heartbeats. Satisfied by filesmgr.Manager.
+// BundleStateRetriever provides read-only access to installed, installing,
+// and failed bundle state for heartbeats. Satisfied by filesmgr.Manager.
 type BundleStateRetriever interface {
 	List() []filesmgr.FileEntry
-	// ListFailures reports bundles whose most recent install attempt failed.
-	// See filesmgr.FailureEntry for why this is in-memory only.
-	ListFailures() []filesmgr.FailureEntry
+	// ListPending reports bundles currently installing or whose most recent
+	// install attempt failed. See filesmgr.FileEntry's doc comment for why
+	// this is in-memory only.
+	ListPending() []filesmgr.FileEntry
 }
 
 type heartbeater struct {
@@ -211,15 +212,16 @@ func (hb *heartbeater) getGroupState() map[string]messages.GroupStateInfo {
 	return gs
 }
 
-// getBundleState reports, per bundle name, both installed and failed state,
-// following the same State+Error pattern as getPolicyState/
-// getBackendState: a name that has successfully installed at least once but
-// whose most recent attempt (e.g. a re-fetch on reconnect) failed reports
-// State as BundleStateFailed with Version updated to the failed attempt's
-// version and Error set, while still retaining SHA256/InstalledAt from the
-// last successful install — mirroring how BackendStateInfo keeps
-// LastRestartTS/LastRestartReason alongside a current Error. A name with no
-// successful install at all reports State/Version/Error from the failure only.
+// getBundleState reports, per bundle name, installed/installing/failed state,
+// following the same State+Error pattern as getPolicyState/getBackendState: a
+// name that has successfully installed at least once but whose most recent
+// attempt (e.g. a re-fetch on reconnect) is in flight or failed reports State
+// as BundleStateInstalling/BundleStateFailed with Version updated to that
+// attempt's version and Error set (when failed), while still retaining
+// SHA256/InstalledAt from the last successful install — mirroring how
+// BackendStateInfo keeps LastRestartTS/LastRestartReason alongside a current
+// Error. A name with no successful install at all reports State/Version/Error
+// from the pending entry only.
 func (hb *heartbeater) getBundleState() map[string]messages.BundleStateInfo {
 	bs := make(map[string]messages.BundleStateInfo)
 	if hb.bundleRetriever == nil {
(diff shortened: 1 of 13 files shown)


leoparente:
@codex review

agent/configmgr/fleet/heartbeats.go
```diff
@@ -21,10 +21,13 @@ const (
...
 
-// BundleStateRetriever provides read-only access to installed bundle state for
-// heartbeats. Satisfied by filesmgr.Manager.
+// BundleStateRetriever provides read-only access to installed and failed
+// bundle state for heartbeats. Satisfied by filesmgr.Manager.
 type BundleStateRetriever interface {
 	List() []filesmgr.FileEntry
+	// ListFailures reports bundles whose most recent install attempt failed.
+	// See filesmgr.FailureEntry for why this is in-memory only.
+	ListFailures() []filesmgr.FailureEntry
```
leoparente: why whole new structure with a lot of duplicated fields instead of merging it in FileEntry?

leoparente: Question: could we have similar states like: Installing, installed, failed, similar what policy state have?

samiura: I have kept them separate mainly because `FileEntry` is persisted to `state.json` and FailureEntry deliberately isn't!! A  failed attempt has no on-disk artifact worth surviving a restart, and every `restart/reconnect` already triggers a fresh install attempt anyway. Merging them would mean either persisting failure fields we don't want persisted, or adding persistence-exempt fields to a type whose whole job today is "what's on disk," which felt like it'd blur that contract. Open to merging if you'd rather have one type — happy to fold it in if you think the split isn't worth it.

samiura: For your second question,

Worth noting `PolicyState` itself doesn't actually have an "in progress" state either — it's Unknown/Running/FailedToApply/Offline, no transitional state. 

But for bundles it's more feasible than it might look: the heartbeat ticks independently every 5s regardless of whether an Ensure call is in flight, so we could set a transient "installing" marker at the start of Ensure and clear it to installed/failed at the end, and the next tick would just pick up whatever's current. 

It's a fairly small addition on top of what's here — do you think we should add it in this PR? or should I file it as a fast-follow?

leoparente: I think worth us having that State, and folding the Failure in the FileEntry, so it would have a state and error field to add or somthing like that. PS: "Running"  for polcies is a transational state in some way

agent/filesmgr/filesmgr.go
```diff
@@ -442,16 +451,68 @@ func (m *filesmgr) List() []FileEntry {
...
-// the declared SHA256. Correct operation order (atomicity guarantee):
+// the declared SHA256. It wraps ensureLocked to record/clear per-name failure
+// state for heartbeat reporting: a failing attempt is recorded via
+// recordFailure so getBundleState (agent/configmgr/fleet/heartbeats.go) can
+// surface it instead of silently omitting the bundle; a succeeding attempt
+// clears any previously recorded failure for the same name.
+func (m *filesmgr) Ensure(ctx context.Context, spec FileSpec) (string, error) {
+	path, err := m.ensureLocked(ctx, spec)
+	if err != nil {
+		m.recordFailure(spec.Name, spec.Version, err, errors.Is(ctx.Err(), context.DeadlineExceeded))
```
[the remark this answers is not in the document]
samiura: addressed with new commit.

agent/configmgr/fleet/heartbeats.go
leoparente: You can fold it in that PR to only increase schema version once

agent/filesmgr/types.go
```diff
@@ -99,28 +99,61 @@ func (s FileSpec) Validate() error {
...
-	FailedAt time.Time `json:"failed_at"`
+	// State, Error, Timeout, and UpdatedAt below are populated only on
+	// entries returned by Manager.ListPending (installing/failed); they are
+	// always zero on entries returned by Manager.List.
+	State string `json:"state,omitempty"`
+	Error string `json:"error,omitempty"`
+	// Timeout is true when a failed attempt's context deadline was exceeded
+	// (the 10-minute install timeout), as opposed to a checksum mismatch or
+	// download/extract error.
+	Timeout   bool      `json:"timeout,omitempty"`
```
leoparente: Timeout   bool      `json:"timeout,omitempty"` I think it is not actually required if we report a proper message in the Error string

leoparente:
@codex review

agent/filesmgr/types.go
samiura: addressed with the latest commit.

leoparente (approved):

PR #471, opened 2026-07-21
feat(snmp-discovery): add emit_device_name option to suppress device name

leoparente:
## Summary

Adds an opt-in `emit_device_name` policy option to snmp-discovery. When set to `false`, snmp-discovery still walks `sysName` but does **not** emit `Device.name` on the matched device, so continual discovery under Assurance stops proposing a hostname rename when the device's `sysName` differs from its NetBox name. **Defaults to `true`** — unset means today's behavior, zero change for existing operators.

Closes ENGHLP-1493 (hostname-omit request).

## Problem

snmp-discovery unconditionally emits `Device.name` from `sysName`. When a policy already matches an existing NetBox device (via a target `netbox_id` / `metadata.source_match`), the match succeeds but the emitted name is still a deviation — so every continual-discovery cycle proposes a hostname rename that never settles. There was no way to keep the match while suppressing the name.

## What changed

- **`config.options.emit_device_name`** (tri-state `*bool`, default `true`), with accessor `DeviceNameEmissionEnabled()` mirroring the existing `emit_prefixes` / `PrefixEmissionEnabled()` pattern.
- **`mapping.ApplyDeviceNameEmission`** — a guarded post-map transform that clears the device name when the option is off.
- Wired into `policy.runWithMetadata` **after** source_match annotation (so the guard can see `netbox_id`) and **before** `PruneNestedRefs` (so the nested device stubs inherit the cleared name).

## The matcher guard (why this is safe)

`Device.name` is a primary device matcher, and interfaces/IPs/MACs reference the device as a **pruned matcher-only stub**. So the name is suppressed only when the device carries a matcher that *also survives onto those nested stubs*:

- ✅ `source_match` (from `netbox_id`) — carried on stubs
- ✅ non-empty `asset_tag` — carried on stubs
- ❌ `serial` — NetBox `Device.serial` is not unique, so it is not a matcher at all
- ❌ `primary_ip` — matches a top-level device, but is dropped from nested stubs (setting it on a stub fails ingest)

If neither qualifying matcher is present, the name is **kept** and a `WARN` is logged, to avoid emitting an unmatchable device that would create a nameless duplicate.

## Virtual-chassis stacks

On a stack the master is also represented by the shared `VirtualChassis.Master` matcher ref. Suppression is applied there too, so the rename is suppressed consistently across every representation of the master (and the `buildMasterRef` no-divergence invariant is preserved). Member names and the virtual-chassis name are untouched.

## Verified

- Unit tests: config accessor; the guard (source_match/asset_tag clear; serial-only and no-matcher keep + WARN; nil/VC-member/blank cases); a virtual-chassis suppression test; a guard↔stub regression test that runs `PruneNestedRefs` after suppression and asserts every nested device stub still carries a matcher; and a WARN-log assertion.
- Live dry-run (snmpsim), device whose `sysName` ≠ NetBox name, checking both the top-level device and nested stubs:
  - `emit_device_name: false` + `netbox_id` → name absent; `source_match.netbox_id` present on device **and** nested stubs; no WARN.
  - `emit_device_name: false` + `defaults.asset_tag` → name absent; `asset_tag` present on device **and** nested stubs; no WARN.
  - `emit_device_name: false`, no matcher → name **kept**; WARN emitted.
  - control (unset) → name present, as today.
- `GOWORK=off go test -race ./...` green; `make fix-lint` → `0 issues.`

## Not included

The originating request also floated deriving `Device.asset_tag` from a base MAC. That was intentionally **not** implemented — `asset_tag` is a physical inventory identifier, not a MAC, and the use case is too narrow to justify the behavior. Hostname omission ships on its own.


leoparente:
@codex review

samiura (approved):

PR #487, opened 2026-07-27
fix(snmp-discovery): complete Cisco sysObjectID model table from upstream MIB

leoparente:
## What

Brings the Cisco `sysObjectID` → model table to **full parity with the upstream CISCO-PRODUCTS-MIB**, adding the **364** entries it was missing (`ciscoProducts` 2046-3542).

Reported in #482: a Catalyst 1200 (`C1200-8FP-2G`) logs

```
{"level":"WARN","msg":"error getting device model falling back to OID","backend":"snmp_discovery",
 "device_oid":".1.3.6.1.4.1.9.1.3213","error":"device ID .1.3.6.1.4.1.9.1.3213 not found"}
```

and the device lands in NetBox with the raw OID as its model. Our table topped out at suffix 3025 with **zero** entries in the 3100-3400 range, so the entire Catalyst 1200/1300 smart-switch family (3210-3253) was missing along with ~300 other recent products.

## How

Generated from **Cisco's own MIB** (`github.com/cisco/cisco-mibs`, `v2/CISCO-PRODUCTS-MIB.my`) and appended only OIDs absent from the file.

Worth noting for future refreshes: the LibreNMS copy of this MIB (26.2.0, `LAST-UPDATED 202410070000Z`) stops at 3383 and is short by 107 products relative to upstream — Catalyst 9350, C8000 G2, Nexus 9364D-GX2A, VG410, UCS X210c and others. Upstream is the better source.

The generator strips ASN.1 comments before matching. `ciscoProducts 229` defines a live `ciscoMGX8830` immediately followed by a commented-out `ciscoMGX8820` alias, and a comment-blind parser silently takes the dead name — there is a test pinning the live one.

After the change: **2949 entries, no duplicates, and no OID present upstream but missing here.**

Deliberately **not** changed: seven pre-existing entries whose historical alias differs from the current MIB object name (`catC930024T` vs `ciscoCat930024T`, `ciscoC9407R` vs `ciscoCat9407R`, the three `*FixedSwitchStack` names, …). Rewriting them would change the model string emitted for devices already discovered into users' NetBox, causing device-type churn for no functional gain. They resolve correctly today.

## Verification

The reporter's OID and the newly-added ranges resolve through the real loader:

```
.1.3.6.1.4.1.9.1.3213 -> ciscoC12008FP2G   (the C1200 from #482)
.1.3.6.1.4.1.9.1.3293 -> ciscoC935048U
.1.3.6.1.4.1.9.1.3542 -> ciscoC8475G2      (highest suffix upstream)
.1.3.6.1.4.1.9.1.229  -> ciscoMGX8830      (live name, not the dead alias)
```

- `TestEmbeddedLookupExtensions_CiscoCatalyst1200And1300` pins the C1200/C1300 family plus the above.
- Every one of the added entries was audited against the upstream MIB's *live* definitions — zero disagreements.
- Existing `_NoDuplicateOIDs` / `_AllOIDsMappedCorrectly` still pass.
- `GOWORK=off go test -race ./...` green across all 12 packages; `make fix-lint` reports 0 issues.

## Note on the second half of #482

The reporter also mentions interface VLANs all showing as access VLAN 1. That is **not** addressed here and does not reproduce: I replayed three real CISCOSB device recordings (`ciscosb_sge`, `ciscosb_cbs350`, `ciscosb_cbs250-24p-4x-v3`) through the VLAN post-pass and the classification is correct — `ciscosb_sge` resolves `{vid 1: 348 ports, vid 181: 44 ports}`, preserving the non-default VLAN. CISCOSB switches omit `dot1qVlanStaticEgressPorts`/`UntaggedPorts` but do populate `dot1qPvid`, which the extractor already handles as its PVID-only signal. Following up with the reporter for their `dot1qPvid` walk before changing anything there.

Refs #482.


samiura (approved):

leoparente:
@codex review

PR #491, opened 2026-07-28
fix(snmp-discovery): correct VLANs on Cisco small-business switches

### LINKED ISSUE #482: Add Support for Cisco C1200 Switches
Hello,

is it possible to add the MIBs of Cisco 1200 Series Switches? We are using a bunch of them in our company and it would be neat if we could use netbox orb to inventory them. The current mibs can be found here: https://software.cisco.com/download/home/286331141/type/286332470/release/4.1.9.85

If you need some more information i'll happily assist.

leoparente:
## What

Cisco small-business switches (Catalyst 1200/1300, CBS/SG) report **every port as access VLAN 1**, whatever the ports are actually configured for. Reported in #482 on a C1200-8FP-2G.

## Why

On these switches the standard Q-BRIDGE sources are not merely absent, they are **wrong**. From the reporter's walks:

- `dot1qPvid` returns **1 for all 14 ports**, including a port genuinely on VLAN 2137.
- `dot1qVlanStaticEgressPorts` / `UntaggedPorts` return **all-zero** masks for all 19 VLANs.

So there is nothing for the generic classifier to correct itself with, and it faithfully reports what the device claims.

The same device populates a private per-port table with the real values, keyed by `ifIndex` rather than bridge port:

```
.1.3.6.1.4.1.9.6.1.101.48.62.1.1.<ifIndex>   vlanAccessPortModeVlanId
.1.3.6.1.4.1.9.6.1.101.48.61.1.1.<ifIndex>   vlanTrunkPortModeNativeVlanId
```

Their actual values — `2128, 2137, 2137, 2137, 2137, 112, 1399, 1399, 1, 100, …` — are exactly what NetBox should show.

## How

Walk those two columns on Cisco-matched hosts and use them for the untagged VLAN.

Gating uses the **existing** `vendor: "cisco"` key, which already covers these devices. Worth calling out: they report `sysObjectID` under **ciscoProducts** (`.1.3.6.1.4.1.9.1.3213`), *not* under the private `.9.6.1` arc, so a prefix test on the private arc would have missed the very device being fixed. Other Cisco platforms do not answer these OIDs, so `HasData()` is false and the overlay never runs.

## Scope: untagged VLAN only

The overlay does **not** derive tagged membership or access-vs-trunk mode, so a trunk on one of these switches is still reported as an access port carrying its native VLAN. The docs state this rather than implying full classification.

I implemented that half and then removed it. `rldot1qPortVlanStaticTable` would supply both, but its columns come back empty in practice, so there is nothing to derive membership or mode from — and the table is expensive to walk, at twelve 128-byte columns per port. `vlanPortModeState` is not a substitute either: a port configured as a trunk reports the same value as its access neighbours.

Re-adding it is a small change if firmware that populates the table turns up.

## Behavioural notes

An access column reporting VLAN 1 **is** honoured, because an operator configured it. The trunk-native column reading 1 is the factory default and indistinguishable from unset, so it is ignored on an access port. The two columns are deliberately not interchangeable, and a test pins that.

Mode and the tagged set are never touched, so a correctly classified trunk cannot be demoted. Ports absent from `infos` are never created, matching `ApplyCisco`.

## Testing

- `qbridge` unit tests: the #482 correction, trunk not demoted, factory-default guard, absent/zero/out-of-range columns leaving the generic result alone, unknown ifIndex ignored, access column outranking native.
- Mapper-level tests through the real `PostMap`, on a fixture modelling the device (wrong `dot1qPvid=1`, empty standard masks, correct private column): untagged becomes **2137, not 1**; a non-CISCOSB host is unchanged; a port the bridge table omits is not mutated; and the two columns are proven distinct.
- `mapping.yaml` guard test, since the collector matches on OID prefixes: dropping a walk entry would silently regress to VLAN 1 with nothing failing loudly.
- Full suite green under `-race`; `make fix-lint` clean.
- Mutation-checked in a scratchpad copy: seven semantic mutants including a no-op control, **all seven killed**.

Fixes #482

### DIFF
--- a/docs/backends/snmp_discovery/supported_platforms.md
+++ b/docs/backends/snmp_discovery/supported_platforms.md
@@ -94,8 +94,13 @@ Switchport-to-VLAN association discovery is built on standard MIBs with one vend
 |-------|----------------|------------------|
 | **Generic** | Q-BRIDGE-MIB (RFC 4363, `1.3.6.1.2.1.17.7.1.4`) + BRIDGE-MIB `dot1dBasePortIfIndex` | VLAN catalog (`dot1qVlanStaticTable` — names + admin status), per-port PVID (`dot1qPvid`), per-VLAN egress + untagged port masks (`dot1qVlanStatic{Egress,Untagged}Ports`). Required for trunk classification. |
 | **Cisco overlay** | CISCO-VLAN-MEMBERSHIP-MIB `vmMembershipTable`, CISCO-VOICE-VLAN-MIB `vmVoiceVlanId` | Access VLAN refinement on non-trunk ports + voice-VLAN promotion. Walked only when `sysObjectID` falls under enterprise prefix `1.3.6.1.4.1.9.` (Cisco Systems) or `1.3.6.1.4.1.29671.` (Meraki). |
+| **CISCOSB overlay** | CISCOSB private `vlan` group (`1.3.6.1.4.1.9.6.1.101.48`): `vlanAccessPortModeVlanId`, `vlanTrunkPortModeNativeVlanId` | Corrects the untagged VLAN on Cisco small-business switches, where the standard sources are wrong rather than absent. Indexed by `ifIndex` rather than bridge port. Shares the Cisco vendor gate above. |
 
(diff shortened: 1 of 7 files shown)


leoparente:
@codex review

samiura (approved):

PR #495, opened 2026-07-29
fix(device-discovery): omit unset device name from nested entity stubs

samiura:
`prune_nested_refs` built nested reference stubs by passing accessor reads into the message constructor: `pb.Device(name=d.name)`. `Device.name` declares explicit proto presence, and reading it while unset yields the proto3 default `""`, which the constructor then records as PRESENT. So a device whose hostname was never discovered was emitted with `name` correctly omitted at top level but `name: ""` on every nested device reference.

The Diode NetBox plugin reads an explicit empty value as a real value. On plugin v1.13.0 it counts as "this field changed" but is stripped from the payload actually written, which is exactly the reported symptom: a deviation claiming one change, showing an empty diff, applying with no effect, and reappearing on the next discovery because nothing was ever written. On v1.14.1 the same payload instead erases the device name in NetBox.

Add `proto_presence.copy_scalar_if_set` and route the presence-bearing scalars through it: `Device.name` in `_device_match_stub`, `Module.serial`, `ModuleBay.position`, and `Device.name`/`serial` on the virtual-chassis master ref in translate_chassis (defensive there — neither can be blank today). Fields without explicit presence stay on the constructor path, since for them empty and unset are indistinguishable on the wire.

### DIFF
--- a/orb-discovery/device-discovery/device_discovery/proto_presence.py
+++ b/orb-discovery/device-discovery/device_discovery/proto_presence.py
@@ -1,0 +1,63 @@
+#!/usr/bin/env python
+# Copyright 2026 NetBox Labs Inc
+"""
+Presence-safe helpers for populating Diode ingester protos.
+
+MOST optional scalars on the ingester protos declare explicit presence (``optional`` in
+proto3) — as a rule, everything except the fields NetBox requires. ``Device`` alone has
+eight such string fields (``name``, ``serial``, ``asset_tag``, ``face``, ``status``,
+``airflow``, ``description``, ``comments``), and ``Interface``, ``IPAddress``,
+``Prefix``, ``VRF``, ``Module`` and ``ModuleBay`` all carry their own. Do NOT work from
+a memorised list: check ``DESCRIPTOR.fields_by_name[name].has_presence``.
+
+For a presence-bearing field, "unset" and "set to the empty string" are DISTINCT on the
+wire, and the Diode NetBox plugin treats them very differently. Unset means "leave this
+field alone"; an explicit empty value is a real value the plugin diffs against NetBox
+and — from plugin v1.14.1 on — writes as a deliberate clear.
+
+The exceptions are the required matcher fields (``Site.name``, ``Interface.name``,
+``DeviceType.model``, ``Manufacturer.name``, ``MACAddress.mac_address``,
+``IPAddress.address``, ``VRF.name``, ...), which have no presence: for them empty and
+unset serialize identically and neither helper here changes anything.
+
+Reading an unset scalar through the normal attribute accessor yields the proto3 default
+(``""``), so the natural-looking ``pb.Device(name=d.name)`` silently launders absence
+into an explicit empty string. See ``copy_scalar_if_set``.
+"""
+
+from google.protobuf.message import Message
+
+
+def copy_scalar_if_set(dst: Message, src: Message, *fields: str) -> None:
+    """
+    Copy scalar ``fields`` from ``src`` to ``dst`` without laundering absence into presence.
+
+    Use this instead of passing accessor reads into a message constructor
+    (``pb.Device(name=d.name)``). When ``d.name`` is unset that expression yields ``""``
+    and the constructor marks the field PRESENT, turning "we discovered no hostname"
+    into "the hostname is empty".
+
+    That laundering is a real reported failure: nested device stubs built by
+    ``prune_nested_refs`` carried ``name: ""`` even when discovery had correctly reported
+    no hostname. On diode plugin v1.13.0 the plugin counted the empty value as a change
+    but stripped it from the payload it actually wrote, so Assurance raised a deviation
+    claiming one change with an empty diff — it applied with no effect and came back on
+    the next discovery. On v1.14.1 the same payload instead erases the device name.
+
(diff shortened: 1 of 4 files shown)


orb-discovery/device-discovery/device_discovery/proto_presence.py
```diff
@@ -0,0 +1,63 @@
...
+def copy_scalar_if_set(dst: Message, src: Message, *fields: str) -> None:
+    """
+    Copy scalar ``fields`` from ``src`` to ``dst`` without laundering absence into presence.
+
+    Use this instead of passing accessor reads into a message constructor
+    (``pb.Device(name=d.name)``). When ``d.name`` is unset that expression yields ``""``
+    and the constructor marks the field PRESENT, turning "we discovered no hostname"
+    into "the hostname is empty".
+
+    That laundering is what produced ENGHLP-1526: nested device stubs built by
```
leoparente: dont mention ticket in the pr

orb-discovery/device-discovery/device_discovery/stubs.py
```diff
@@ -179,7 +180,11 @@ def _device_match_stub(
     defaults.device.asset_tag — kept on the stub so the rich entity and stub never resolve
     via different matchers.
     """
-    stub = pb.Device(name=d.name)
+    # copy_scalar_if_set, not pb.Device(name=d.name): Device.name declares proto
+    # presence, so reading an unset name as "" and passing it to the constructor
+    # would mark the field PRESENT and send an explicit empty name (ENGHLP-1526).
```
leoparente: +1

samiura: Done — removed from all code comments and both commit messages, force-pushed. Also worth noting tests/custom_drivers/cisco_ios/test_driver.py:55 has an ENGHLP-1279 reference on develop if you want that can be cleaned up separately.

samiura: https://github.com/netboxlabs/orb-agent/pull/499

leoparente:
Can you explain the solution here? because device name is never empty as it uses hostname

samiura:
> Can you explain the solution here? because device name is never empty as it uses hostname

You're right, and my description overstated it — sorry for the confusion.

I have checked all 41 drivers: every one falls back to `self.hostname` or `"Unknown"`, so `device.name` is never empty on those paths. 

The one exception is `nokia_sros.py:932` — `_find_txt` defaults to `""` on an XPath miss with no fallback, so a NETCONF reply missing `oper-name` gives `hostname = ""`. But that's an empty string, not an unset field, so #496 is what fixes it, not this PR.

The actual reason this one is needed: #496 doesn't work without it. Same input (`hostname=""`):

| | top-level `device.name` | nested `device.name` |
|---|---|---|
| #495 only | present, `""` | present, `""` |
| #496 only | absent ✅ | **present, `""`** ❌ |
| both | absent ✅ | absent ✅ |

`blank_to_none` correctly unsets the name, then `prune_nested_refs` reads that unset field as `""` and re-materialises it as *present* on every nested device ref. So #496 alone fixes the device entity and leaves every interface still carrying `name: ""`.

So this isn't an independent bug fix — it's the precondition that makes #496 hold end-to-end. Happy to fold the two together into one PR if that's easier to review, since neither does much alone.

leoparente:
Wouldnt make sense to fallback on our side then if hostname is empty for us to use self.hostname instead of applying blank? Then we can add a this feature https://github.com/netboxlabs/orb-agent/commit/a7c898d5e76641c4791ac38edbc0e2ef4d9f35a7 so avoid emiting the name when asset_tag and/or source_id exists

samiura:
> Wouldnt make sense to fallback on our side then if hostname is empty for us to use self.hostname instead of applying blank? Then we can add a this feature [a7c898d](https://github.com/netboxlabs/orb-agent/commit/a7c898d5e76641c4791ac38edbc0e2ef4d9f35a7) so avoid emiting the name when asset_tag and/or source_id exists

Both make sense.

**nokia_sros** — agreed, fixed at source in #496: `hostname = _find_txt(...) or self.hostname`, matching the other 40 drivers. `fqdn` derives from the same value so that stops being blank too. Added a `missing_oper_name` get_facts scenario that fails without it.

Keeping `blank_to_none` as a backstop though, since hostname isn't the only field it covers — `Interface.description` (17 drivers hardcode `"description": ""` without ever reading one off the box), `Device.serial`, and `asset_tag` / `VRF.rd` from policy.

**emit_device_name** — that's a better argument for #495 than the one I gave. Your Go version does `dev.Name = nil`, which is safe because `diode-sdk-go` has `Name *string`. Python protobuf has no equivalent: reading an unset `optional string` returns `""`, and assigning it back marks the field *present*. So porting it to device-discovery without #495 would have `prune_nested_refs` re-materialise `name: ""` on every nested ref — an unmatchable device, which is what `deviceHasAlternativeMatcher` guards against. `_master_device_ref` here is the Python counterpart of your `clearVirtualChassisMasterName`.

Happy to port emit_device_name as a follow-up once these land.

leoparente (approved):
Please, do the followup pr of emit_device_name and remeber to update orb-agent docs

leoparente:
@codex review

PR #496, opened 2026-07-29
fix(device-discovery): omit blank values instead of sending empty strings

samiura:
Follows #495, which landed the `copy_scalar_if_set` half. Two commits: the blank-value handling, and the nokia_sros fallback Leo asked for in review.

## Problem

Two sources feed blank strings onto presence-bearing proto fields, where the Diode plugin reads them as real values rather than omissions.

**Drivers.** NAPALM's getter contract uses `""` for *"this value is not available"*, not for *"this value is empty on the device"*. 17 of the custom drivers hardcode `"description": ""` in `get_interfaces` without ever reading a description off the box:

```
$ grep -rl '"description": ""' custom_napalm/ | wc -l
17
```

A driver that cannot determine a hostname or serial can likewise return `""` rather than `None`.

**Policy.** A policy may legitimately be written with an empty `asset_tag` or VRF `rd`.

## Why the policy fields matter more than the empty diff

`asset_tag` is the **highest-precedence** `dcim.device` matcher. And the nested stubs already omit a falsy `asset_tag` / `rd` (`stubs.py:191`, `stubs.py:31`) while the rich entity carried the empty value — so a blank policy value made the rich entity and its own stub resolve through **different matchers**. That directly violates the invariant `_device_match_stub`'s docstring declares:

> `asset_tag` is the highest-precedence matcher and is populated when the policy sets `defaults.device.asset_tag` — kept on the stub so the rich entity and stub never resolve via different matchers.

There is a test for exactly this consistency property.

## Change

Adds `proto_presence.blank_to_none` and applies it where driver or policy values reach presence-bearing fields:

| field | source | file |
|---|---|---|
| `Device.name` | driver `hostname` | `translate.py` |
| `Device.serial` | driver `serial_number` | `translate.py` |
| `Device.asset_tag` | policy | `translate.py` |
| `VRF.rd` | policy | `translate.py` |
| `Interface.description` | driver | `interface.py` |
| `Interface.mac_address` | driver | `interface.py` |
| `Module.serial`, `ModuleBay.position` | driver payload | `translate_modules.py` |

Details worth a look in review:

- The `serial_number` call sits **after** the existing list/tuple normalisation, so `[""]` is caught too. Placing it earlier would miss that case.
- Blank means blank-or-whitespace. Whitespace-only values only ever come from a parser that matched nothing. Non-blank values pass through **unstripped**, so genuine content is never rewritten.
- `bytes` is handled alongside `str`, because `translate_device` deliberately lets a `bytes` serial through its type normalisation.
- The pre-existing `mac_address != ""` special case becomes a call to the shared helper — same behaviour, plus whitespace.
- `Module.serial` / `ModuleBay.position` are defensive: `custom_napalm/_modules.py` drops blank-serial bays today, so these only fire if a driver assembles the envelope without `to_payload`.

## Two deliberate behaviour changes — please confirm these are the calls you want

**1. Discovery can no longer clear a description an operator set in NetBox.** Under NAPALM's contract a blank getter value is not evidence the field should be cleared, and on plugin v1.13.0 such writes were stripped anyway — so this matches current effective behaviour. But on v1.14.1 it forgoes a capability the plugin now offers.

**2. A policy setting `defaults.interface.description` now applies on the 17 drivers that hardcode a blank description**, where previously the blank won and overrode the policy default with `""`.

Open question: would you rather the description handling be opt-in via a policy option than a blanket rule? I went with the blanket rule because a per-driver audit of which drivers genuinely read descriptions is a much larger change, but it is a product call.

One further silent change, smaller: a driver returning `"description": None` previously overrode the policy default with `None`; it now keeps the default.

## Testing

`tests/test_blank_napalm_values.py` — 20 tests, including a parametrised sweep asserting that **no** emitted entity carries a presence-bearing string field that is present but blank, across 11 input shapes (blank/whitespace/None hostname, blank/whitespace/list/bytes serial, blank/whitespace description, blank MAC, everything-at-once).

Against unmodified `develop`, 14 of the tests in this file fail. The sweep's blank check is `not value.strip()` rather than `== ""` specifically so the whitespace cases actually bite — with an equality check they passed both before and after and were worthless.

Full backend suite: 1880 passed, 161 skipped. `ruff check .` clean.

## Risk

Moderate, concentrated in the two behaviour changes above. Everything else replaces a payload the plugin either could not apply or applied destructively.

`_present_but_blank` in the test file probes for `FieldDescriptor.is_repeated` and falls back to `.label`, since `is_repeated` landed in protobuf 5.27 and `.label` was removed in protobuf 6 — the backend does not pin protobuf, and CI runs Python 3.11–3.14.

**nokia_sros fallback** (second commit) — `get_facts` used `_find_txt(...)` directly, which returns `""` on an XPath miss, so a NETCONF reply without `state/system/oper-name` gave a blank hostname and a blank `fqdn`. Now falls back to `self.hostname`, matching the other 40 drivers. New `missing_oper_name` get_facts scenario — the existing mock with `<oper-name>` removed — fails without the change.

### DIFF
--- a/orb-discovery/device-discovery/device_discovery/translate.py
+++ b/orb-discovery/device-discovery/device_discovery/translate.py
@@ -32,6 +32,7 @@ from device_discovery.policy.models import (
     TenantParameters,
     VrfParameters,
 )
+from device_discovery.proto_presence import blank_to_none
 from device_discovery.translate_chassis import (
     translate_as_stack,
     validate_chassis_payload,
@@ -72,7 +73,10 @@ def translate_vrf(
     if isinstance(vrf, VrfParameters):
         return VRF(
             name=vrf.name,
-            rd=vrf.rd,
+            # VRF.rd declares presence, and stubs.py omits a falsy rd — a policy
+            # supplying rd="" would otherwise make the rich VRF and its stub resolve
+            # via different matchers.
+            rd=blank_to_none(vrf.rd),
             comments=vrf.comments,
             description=vrf.description,
             tags=vrf.tags,
@@ -141,6 +145,9 @@ def translate_device(
                 serial_number = str(serial_number[0])
     elif serial_number is not None and not isinstance(serial_number, str | bytes):
         serial_number = str(serial_number)
+    # Device.serial declares proto presence too, so a blank serial must be omitted
+    # rather than sent as an explicit empty value.
+    serial_number = blank_to_none(serial_number)
 
     device_config = None
     if config_info and options:
@@ -148,12 +155,19 @@ def translate_device(
 
     # Build Device parameters
     device_params = {
-        "name": device_info.get("hostname"),
+        # A driver that discovered no hostname must OMIT device.name. An explicit
+        # empty name is a real value to the Diode plugin, not an omission.
+        "name": blank_to_none(device_info.get("hostname")),
         "device_type": DeviceType(model=model, manufacturer=manufacturer),
         "platform": Platform(name=platform, manufacturer=manufacturer),
         "role": defaults.role,
         "serial": serial_number,
-        "asset_tag": defaults.device.asset_tag if defaults.device else None,
+        # asset_tag is the highest-precedence dcim.device matcher and comes straight
+        # from policy. Blanking it here keeps the rich entity consistent with the
+        # nested stub, which omits a falsy asset_tag.
+        "asset_tag": blank_to_none(defaults.device.asset_tag)
+        if defaults.device
+        else None,
         "status": "active",
         "site": defaults.site,
         "tags": tags,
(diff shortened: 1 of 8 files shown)


orb-discovery/device-discovery/device_discovery/translate.py
```diff
@@ -141,19 +145,29 @@ def translate_device(
...
 
     device_config = None
     if config_info and options:
         device_config = translate_device_config(config_info, options)
 
     # Build Device parameters
     device_params = {
-        "name": device_info.get("hostname"),
+        # A driver that discovered no hostname must OMIT device.name. An explicit
+        # empty name is a real value to the Diode plugin (ENGHLP-1526).
```
leoparente: dont mention ticket in the pr in OSS

leoparente:
@codex review

leoparente (approved):

PR #506, opened 2026-08-05
feat(device-discovery): add emit_device_name to suppress the discovered hostname

leoparente:
## What

device-discovery always emits `Device.name` from the hostname the driver reported. When a policy matches an existing NetBox device by `netbox_id` and the discovered hostname differs from the NetBox name, the match succeeds but the emitted name is a deviation — so continual discovery proposes a hostname rename on every cycle and it never settles. There was no way to keep the match and drop the name.

Requested as a device-discovery equivalent of the snmp-discovery option added in a7c898d (#471). This ports it, including the matcher guard and the master-only scope on virtual-chassis stacks.

Default is `true`, so existing policies are unaffected.

```yaml
policies:
  my-policy:
    config:
      options:
        emit_device_name: false
    scope:
      - driver: ios
        hostname: 10.0.0.5
        netbox_id: 1234        # or defaults.device.asset_tag
```

## The part that would have been a silent data-loss bug

`Device.name` declares **explicit proto presence**, and this repo's own `proto_presence.py` documents that for such a field the Diode plugin treats an explicit empty string as a *deliberate clear* from plugin v1.14.1 on.

So `device.name = ""` would not suppress the name — it would **wipe the hostname in NetBox**, which is the exact opposite of the intent. Suppression uses `ClearField`. A test asserts `HasField("name") is False`, and reverting to `""` fails three tests.

## Why the guard refuses to fire without an alternative matcher

`name` is a primary NetBox device matcher. Dropping it unguarded emits a device NetBox cannot resolve, which is worse than the churn being fixed. So suppression only happens when the device carries a matcher that **also survives onto the nested device stubs**:

- `metadata.source_match` — written from a scope `netbox_id`
- a non-empty `asset_tag` — from `defaults.device.asset_tag`

Otherwise the name is kept and a warning is logged.

Two exclusions are deliberate and both are tested:

| Field | Why it does not qualify |
|---|---|
| `serial` | NetBox `Device.serial` is not unique and generates no matcher at all. The test fixture always carries a serial, so a test pins that serial alone does **not** unlock suppression. |
| `primary_ip4` / `primary_ip6` | `unique_primary_ip4`/`ip6` match a top-level device, but the nested stubs drop primary_ip (setting it on a nested stub fails ingest), so clearing the name would leave every nested reference unmatchable. |

**Practical consequence worth stating plainly:** this option only helps operators who match by `netbox_id` or set `asset_tag`. Anyone relying on name-based matching gets the warning and keeps the name, by design.

## Two wiring points, both order-sensitive

snmp-discovery has one emission path; device-discovery has two, and each has an ordering constraint that is easy to get wrong and invisible if you do.

**Standalone** — before `copy.deepcopy(device)` in `translate.py`. Every nested interface and IP reference is derived from that copy, so suppressing after it leaves them carrying the name that was just dropped.

**Stack** — before `_master_device_ref` derives the shared VC master ref. That helper copies the name through `copy_scalar_if_set`, which skips an unset field, so suppressing first means the ref inherits the suppression. Suppressing after resurrects the name that `_master_device_ref` exists to keep in sync with the master. Moving the call one line later fails a test.

That second point came out cleaner than the snmp implementation, which needed a separate `clearVirtualChassisMasterName` pass to fix up refs after the fact. Correct ordering makes it fall out for free here.

## Scope on stacks: master only

Matching snmp-discovery. The transform no-ops on any device carrying `vc_position`, so member names are untouched — they come from `stack_member_name_template`, which is their own lever.

This is a deliberate limitation, not an oversight. Suppressing member names would need a per-member alternative matcher, and members are matched by serial, which is not a unique NetBox matcher. A policy has no way to set `netbox_id` per stack member today. If member-name churn turns out to be the real problem for a reporter, that needs its own design.

## Verification

End-to-end through `translate_data` on both the standalone and 2-member stack paths, asserting the master and the shared VC master ref are both suppressed while member names survive. Full numbers and the mutation matrix are in the review section below, which supersedes them after the fixes in `2778b6f`.

---

## Adversarial review pass

A 5-lens adversarial review (each finding independently refuted before it counted) raised 15 findings; **5 survived**, and they collapse to two distinct defects. Both are fixed in `2778b6f`.

**1. Suppression could disable `prune_nested_refs` entirely.** Four lenses found this independently.

`_resolve_device` only ever tried name, then serial. Suppression is permitted on the strength of `source_match` or `asset_tag` — neither of which pruning consulted. So a device with a scope `netbox_id` and no driver-reported serial resolved to nothing:

```
emit=False, serial present -> ['device_type','metadata','role','site']   properly stubbed
emit=False, NO serial      -> [...,'platform',...,'status']              un-pruned rich copy
                              + "could not resolve nested device ''"     one WARNING per interface
```

`platform` and `status` are fields `_device_match_stub` exists to strip, so their presence proves the ref never went through pruning. Serial-less devices are a real shape, not synthetic: napalm's `iosxr` and `eos` drivers can both yield an empty serial, and `translate_device` blanks it to unset.

Fixed by adding `source_match` and `asset_tag` as resolution fallbacks after name and serial, so pruning resolves in exactly the cases suppression is allowed. Both require a **unique** match, so an ambiguous index is left unresolved rather than mis-attributed to the wrong device.

One caveat on scope: a reviewer also claimed this drops `device.primary_ip4` in NetBox by breaking the primary-IP cycle-closer. I could not reproduce that — in my payload the top-level `primary_ip4` was unset in the baseline too, so my repro never exercised the cycle-closer. The un-pruned-ref and warning-per-interface effects are measured; the primary-IP claim is not, and I have not repeated it as established. The fix removes the precondition either way.

**2. The standalone call-site ordering had no test at all.** Moving `apply_device_name_emission` after `translate.py`'s `copy.deepcopy(device)` previously broke nothing — even though that ordering is the entire reason the call sits where it does. The first commit asserted the constraint in a comment and a commit message without testing it; only the *stack* ordering was pinned. Now both are.

**Final verification.** 1977 passed, 161 skipped, ruff clean. Five mutants each fail a test:

| Mutant | Caught |
|---|---|
| `name = ""` instead of `ClearField` | ✅ |
| matcher guard removed | ✅ |
| stack call moved after `_master_device_ref` | ✅ |
| standalone call moved after the `deepcopy` | ✅ |
| `source_match` / `asset_tag` fallbacks removed | ✅ |

Note `stubs.py` is shared by every translate path, so the fallback is additive by design: it only runs after name and serial have both failed, which previously returned `None` and skipped pruning.


leoparente:
@codex review

samiura (approved):

PR #511, opened 2026-08-05
fix(snmp-discovery): keep ModuleType on nested Module references

leoparente:
## What

`PruneNestedRefs` stripped `ModuleType` from the nested `Interface.Module` ref. That is right for *matching* — `dcim.module` resolves on `module_bay`, its only unique matcher besides `asset_tag` — but a nested reference that has not yet resolved is **created** rather than matched, and NetBox rejects a Module without one:

```
ERR_OPS_GENERATE_DIFF - {"dcim.module":{"module_type":["Field module_type is required"]}}
```

Companion to #509, which fixes the identical defect in device-discovery.

## Why emitting modules first does not already cover this

The old comment reasoned that the ref would resolve because module entities are spliced ahead of interfaces. That does not hold: **`bulk-plan-apply` batches plans and applies them independently, so payload order is not reconciliation order.** I measured this separately while investigating the device-discovery case — reordering a payload so VLANs preceded the interfaces referencing them changed nothing (44 missing vs 36-49 baseline); only a real barrier between two ingests helped, and the agent cannot observe reconciliation completion.

And the blast radius is not confined to the module: a failed plan takes its whole batch down, including unrelated interfaces. In device-discovery this cost 48 failed plans per run on a 6-linecard chassis and **36-49 of 432 interfaces**; fixing it took that to 14, and to 0 combined with the plugin-side fix. This backend has the same stub shape and had simply never been measured on a modular chassis.

## Why only `(Manufacturer, Model)` is copied

`stub.ModuleType = mt` would work, but this stub is duplicated into **every interface on the linecard**, so it must not grow as drivers start populating `part_number`, `description`, `comments`, or `Manufacturer.slug`. `(Manufacturer, Model)` is the complete `dcim.moduletype` matcher and the only pair required on create, so nothing is lost. Matches how the rest of this stub is built field-by-field, and mirrors #509.

## Two comment corrections

Both were factually wrong and would mislead the next person trimming this stub:

- It claimed the ref resolves via *"the (Device, ModuleBay) or (Device, Serial) match paths"*. Per the plugin's matcher table, `dcim.module` has **no serial matcher** — `serial` resolves nothing and is carried for create fidelity only. `module_bay` is the load-bearing field.
- The clear-path comment justified dropping the ref because *"we also strip ModuleType"*. The actual reason is that NetBox requires `module_bay`, which retaining `ModuleType` does not supply. The drop stays correct, for the correct reason.

## gnmi-discovery deliberately not included

Checked, and it needs no equivalent change: gnmi **never attaches a Module to an Interface**. Its only reference to the field is `ifaceCopy.Module = nil` in `translate_ip.go:269`, and there is no assignment anywhere that sets it. Its top-level `Module` entities already carry a manufacturer-bearing `ModuleType` (`translate.go:601`), and its own `case *diode.Module` in `stubs.go` only stubs the nested Device, never touching ModuleType. So there is no nested module ref to fix.

(One PR per backend, per repo convention, hence separate from #509.)

## Verification

- Full snmp-discovery suite green across all 12 packages; `gofmt` clean; `golangci-lint` (v2.11.4 config) **0 issues**.
- Mutation-tested, both caught:

| mutant | caught by |
|---|---|
| ModuleType copy removed entirely | 4 tests, incl. `StubsModuleBayAndInterfaceModule` |
| unbounded `stub.ModuleType = mt` | `InterfaceModuleTypeCopyIsBounded` |

- New test `TestPruneNestedRefs_InterfaceModuleTypeCopyIsBounded` pins that `part_number`, `description`, `comments` and `Manufacturer.slug`/`description` do **not** ride along.
- The three existing assertions that pinned `ModuleType == nil` are inverted, and their prose updated to say why it is retained.

Not verified end-to-end against a live modular chassis over SNMP — I have no such capture. The device-discovery equivalent was validated e2e on a real 432-interface payload; this is the same defect in the same stub shape.

### DIFF
--- a/orb-discovery/snmp-discovery/mapping/stubs.go
+++ b/orb-discovery/snmp-discovery/mapping/stubs.go
@@ -349,31 +349,36 @@ func PruneNestedRefs(entities []diode.Entity, currentDevice *diode.Device, prima
 				e.Lag = stubForIface(e.Lag)
 			}
 			if e.Module != nil {
-				// Reduce nested Interface.Module to a matcher-only ref:
-				// chassis Device stub + Serial (if known) + ModuleBay
-				// matcher (if known). The top-level Module entity
-				// carries the full record (module_type, description,
-				// status, etc.); this nested form lets the Diode
-				// reconciler resolve the ref to that top-level row via
-				// the (Device, ModuleBay) or (Device, Serial) match
-				// paths without re-creating it.
+				// Reduce nested Interface.Module to a matcher-only ref: chassis
+				// Device stub + ModuleBay matcher + ModuleType, plus Serial when
+				// known. The top-level Module entity carries the full record
+				// (description, status, etc.).
 				//
-				// Shape mirrors device-discovery's _module_match_stub
-				// (device-discovery/device_discovery/stubs.py
-				// `_module_match_stub`): emit Device unconditionally,
-				// then conditionally copy Serial and ModuleBay matcher
-				// fields when present on the rich Module. Vendors that
-				// omit transceiver serial in ENTITY-MIB (some Aruba
-				// and low-end OEMs) still populate the bay, so the
-				// (Device, ModuleBay) path alone must keep the ref
-				// resolvable.
+				// ModuleBay is the load-bearing part: it is dcim.module's only
+				// unique matcher besides asset_tag. Serial is carried for create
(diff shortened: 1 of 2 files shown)


jajeffries (approved):
Overall looks good. There's a vulnerability that's probably worth addressing separately.

leoparente:
@codex review

PR #513, opened 2026-08-06
feat(snmp-discovery): support SNMPv3 context name

samiura:
## Summary

SNMPv3 devices that expose their MIB data in a named context (e.g. multi-function printers using `snmpwalk -v3 ... -n <context>`) returned zero entities, because the backend never set a context name on the v3 client — it always queried the default context. The walk succeeds and logs `entity_count=0` with no error, so the failure was silent.

This adds a `context_name` authentication field (SNMPv3 only) and wires it through to the client, so operators can configure `snmpwalk -n`-equivalent behavior instead of falling back to SNMPv1.

- Add `context_name` to `Authentication` (policy-level and per-target) and set it on the gosnmp v3 client's `ContextName` field.
- Support `${VAR}` environment variable substitution for `context_name`, consistent with the other auth fields.
- Reject `context_name` on SNMPv1/v2c with a clear validation error (context is a v3-only concept), including for a `scope.authentication` block that sets `context_name` without `protocol_version` — a case the existing validation gate otherwise skipped entirely.
- Document the new field in the backend README, the config sample, and the env-var substitution table, including a note that per-target authentication overrides the policy-level block wholesale (no field-level merge), so a policy-wide `context_name` is a no-op for any target with its own `authentication` block.

## Test plan

- [x] `go test -race ./...` passes in `orb-discovery/snmp-discovery`
- [x] New unit tests cover: client wiring (including a wire-level encoder check confirming the context name is actually present in the encoded SNMPv3 packet), YAML parsing at both policy and target level, env var resolution (success + missing var), and rejection on SNMPv1/v2c (including the no-`protocol_version` edge case)
- [x] Manual verification against a real or lab SNMPv3 agent configured with a named context (e.g. via `proxy -Cn` / AgentX), confirming `entity_count > 0` with `context_name` set vs. `0` without it

### DIFF
--- a/docs/backends/snmp_discovery/README.md
+++ b/docs/backends/snmp_discovery/README.md
@@ -180,10 +180,11 @@ Each target in the `targets` list can include:
 | auth_passphrase | string | no | SNMPv3 authentication passphrase |
 | priv_protocol | string | no | SNMPv3 privacy protocol (see [SNMPv3 auth/priv protocols](#snmpv3-authpriv-protocols)) |
 | priv_passphrase | string | no | SNMPv3 privacy passphrase |
+| context_name | string | no | SNMPv3 context name (equivalent to `snmpwalk -n`). Required by devices that expose MIB data in a named context; such devices return an empty walk when it is omitted. Rejected for SNMPv1/v2c. |
 
 *Required for SNMPv1/v2c, optional for SNMPv3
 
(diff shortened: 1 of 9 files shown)


leoparente:
@codex review

leoparente (approved):

PR #523, opened 2026-08-10
docs(agent): add Grafana flow monitoring guide

samiura:
Adds `docs/observability/grafana-flow-monitoring.md` — a customer-facing guide for monitoring NetFlow v5/v9, IPFIX, and sFlow in Grafana using orb-agent's pktvisor backend.

## What's covered

- Capability overview (top talkers, ports, conversations, protocol mix, GeoIP, per-direction + packet counters, interfaces, ASN, NetBox device/interface enrichment)
- Install + config (`agent.yaml` flow policy, taps, enrichment, ASN summarization)
- **Recommended path: OTLP push to Grafana Alloy** (no Prometheus scrape needed), with Prometheus scrape kept as a documented alternative for teams already running it
- NetBox enrichment workflow (device_map generation, ifIndex mapping, auto-refresh)
- Dashboard import, full metrics reference, troubleshooting
## Testing

Full pipeline (orb-agent → OTLP → Alloy → Mimir → Grafana) verified end-to-end against a local Docker host — confirmed metric names, dashboard rendering, and OTLP export in logs.

Grafana Cloud remote-write swap is documented but not yet live-tested against a real Cloud stack — pending resolution of a permissions issue tracked in ITH-2245. Will update this PR once that's confirmed.

---
_Note: originally under `docs/integrations/`; moved to `docs/observability/` during review to avoid colliding with the established Discovery Integrations docs category (netboxlabs.com/docs/integrations/), which covers the opposite direction (syncing infrastructure data into NetBox)._


docs/integrations/grafana-flow-monitoring.md
```diff
@@ -0,0 +1,227 @@
...
+Monitor **NetFlow v5/v9, IPFIX and sFlow** in Grafana using **orb-agent**. Includes
+per-direction / packet / interface / ASN metrics and authoritative **NetBox**
+device/interface enrichment.
+
+## Overview
+
+Network devices export flow records to orb-agent, which computes streaming summaries — top
+talkers, ports, conversations, protocol mix, geo, ASN, interfaces, by bytes *and* packets —
+and exposes them as Prometheus metrics for Grafana. Optionally it enriches each exporter and
+interface with its real name from NetBox.
```
leoparente: > Optionally it enriches each exporter and interface with its real name from NetBox.

how it does it? is it true already?

docs/integrations/grafana-flow-monitoring.md
```diff
@@ -0,0 +1,227 @@
...
+
+- Docker or Podman. Images are multi-arch (amd64 + arm64) — runs native on Apple Silicon.
+- Grafana + Prometheus (self-hosted or Grafana Cloud).
+- Devices able to export NetFlow/IPFIX (UDP 9995) or sFlow (UDP 6343).
+- Optional: a **NetBox** instance + API token (enrichment); a GeoLite2-ASN/City `.mmdb`
+  (ASN / geo).
+
+## Install
+
+Create `agent.yaml`:
```
leoparente: I think the example should already contain the OTLP part

docs/integrations/grafana-flow-monitoring.md
```diff
@@ -0,0 +1,227 @@
...
+
+## Configure your exporters
+
+Point devices at `<collector>:9995` (NetFlow/IPFIX) or `:6343` (sFlow). No hardware? On any
+Linux host, `softflowd -i eth0 -n <collector>:9995 -v 10` exports real IPFIX. Confirm flows
+are arriving with `flow_records_flows`.
+
+## Send metrics to Grafana
+
+**Recommended: push OTLP to Grafana Alloy.** orb-agent exports metrics over OTLP straight
```
leoparente: shouldnt we consider this pre required and be merged in the Install session?

docs/observability/grafana-flow-monitoring.md
```diff
@@ -0,0 +1,227 @@
...
+scrape_configs:
+  - job_name: orb-agent
+    metrics_path: /api/v1/policies/__all/metrics/prometheus
+    static_configs:
+      - targets: ["<collector-host>:10853"]
+```
+
+## NetBox enrichment (optional, recommended)
+
+Turns raw exporter IPs into authoritative names (`edge-router-01 / xe-0/0/0`) sourced from
```
leoparente, line 184: where this device_map is supported? dont get it. also this enrichment part will be part of the grafana netbox datasource and not here probably

samiura: `device_map/enrichment` are real fields on `pktvisor`'s `flow_stats` handler config (same block as summarize_ips_by_asn) — tested end to end here against `pktvisor 4.5.0-develop-d942726` under orb-agent, with `device/device_interface` labels resolving correctly on the metrics.

samiura: Good call on the NetBox datasource — I didn't realize when writing the doc. This section now says explicitly that device_map is the collector-layer mechanism available today, and that this doc will get updated to reference the datasource plugin if you think it is worth pointing to.

docs/integrations/grafana-flow-monitoring.md
samiura: Good catch — that line sounds as aspirational. It's real: we generate a `device_map` from NetBox (exporter IP → device name, SNMP ifIndex → interface name) and paste it into the flow policy. Verified end-to-end locally. Updated the doc to say so explicitly right at that line, with a pointer down to the full mechanism.

docs/integrations/grafana-flow-monitoring.md
samiura: Same as the previous reply above.

docs/integrations/grafana-flow-monitoring.md
samiura: Agreed — folded the OTLP config into the Install example directly and removed the separate "Send metrics to Grafana" section. Prometheus is now a clearly-labeled alternative further down for people not introducing Alloy.

docs/observability/grafana-flow-monitoring.md
leoparente: oh I forgot about that.. cool! we should not mention datasource right now

samiura: Okay, got rid of the datasource referrences for now and will wait for it whenever we are ready on that front.

leoparente:
Naming integrations may cause confusion with controller integrations

samiura:
> Naming integrations may cause confusion with controller integrations

Let me think about it, I do not have an immediate answer to this.

@leoparente, By controller integrations, Can you elaborate a bit more whether rename/restructure occurs in the Fleet/Console Integrations settings page, or something else? Happy to move this out of docs/integrations/ once I know what it'd collide with.

The reason for this ask, I do not have the full picture of the controller integration side but happy to investigate more. If it does collide with the naming, what other options we have?

leoparente:
https://netboxlabs.com/docs/integrations/

samiura:
> https://netboxlabs.com/docs/integrations/

Great catch!! That is a real collision, netboxlabs.com/docs/integrations/ is the established "Discovery Integrations" category (Cisco ACI, Meraki, vCenter, etc. syncing into NetBox), which is the opposite direction of what this doc covers. Moved it to docs/observability/ instead — matches the "More NetBox observability with orb-agent" section already in the doc.

leoparente (approved):

PR #524, opened 2026-08-10
ci(repo): suppress pip-vendored CVE findings in orb-agent image scans

unlisted:
## Summary

- Added a root `.trivyignore.yaml` covering three findings that come from pip's *vendored* packages under `pip/_vendor/`, not from any orb-agent dependency: `GHSA-6v7p-g79w-8964` (msgpack 1.1.2, HIGH), `CVE-2025-47273` (setuptools 70.3.0, HIGH), `CVE-2026-59890` (setuptools 70.3.0, MEDIUM). These are what currently block Build & Scan on the release PR #521.
- Wired the ignore file into `build.yaml`, and into `container-rescan.yaml`, which also needed an `actions/checkout`: it only pulled the image, so a repo-root ignore file could never have applied there.
- Added a comment at the pip pin in `agent/docker/Dockerfile`: the ignore entries are pinned by exact PURL, so moving that pin makes them stop matching and the scan fail until they are re-verified.

The diff is additive only. No existing step or script is modified.

### Why suppress rather than fix

pip 26.2.1 is the latest release and vendors exactly these versions, so there is no upgrade path, and the image needs pip at runtime for the entrypoint's `INSTALL_DRIVERS_PATH` / `INSTALL_WORKERS_PATH` feature.

- setuptools (both CVEs): pip vendors only the `pkg_resources` shim. There is no `pip/_vendor/setuptools/` directory and no `package_index.py` anywhere under `_vendor/`. The vulnerable code is absent from the image.
- msgpack: the out-of-bounds read requires reusing an `Unpacker` after a caught exception. pip's only msgpack consumer is vendored CacheControl, which unpacks HTTP-cache entries it serialized itself, with a fresh unpack per read.

Each entry carries a `statement` and `expired_at: 2026-11-08`, so the suppression cannot silently become permanent. Entries are pinned by PURL rather than bare ID so they cannot quietly cover a future, genuinely-reachable copy of the same package.

### What triggered this

No code change here caused it. `python:3.14-alpine` refreshed its bundled pip to 26.2.x, and pip 26.2 is the first release to ship an embedded CycloneDX SBOM at `pip/_vendor/bom.cdx.json` (absent in 25.3/26.0/26.1/26.1.2, present in 26.2/26.2.1). Vendored packages carry no dist-info metadata, so Trivy had no way to see them before that file existed. The vendored versions themselves have not changed since at least pip 25.3, and all three advisories were published well before the alerts appeared on 2026-08-05.

## Test plan

- [x] Scanned the published `netboxlabs/orb-agent:develop` image with Trivy 0.70.0 (the version this action installs): 3 findings before, 0 after.
- [x] Ran the existing `build.yaml` summary script under node against that scan output: `has_blocking = false`.
- [x] Backdated `expired_at` and re-scanned: all three findings return, so the expiry is live rather than decorative.
- [x] Replayed `trivy-action`'s `entrypoint.sh` at the pinned SHA with a stub `trivy` on PATH: resolves to `TRIVY_IGNOREFILE=.trivyignore.yaml`.
- [x] `trivy convert` from the suppressed report yields 0 SARIF results and 0 SBOM vulnerabilities, so the three open code-scanning alerts should close on the first `develop` analysis after merge.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -87,6 +87,9 @@ jobs:
           ignore-unfixed: true
           exit-code: "0"
           list-all-pkgs: "true"
+          # Trivy only auto-detects a plain .trivyignore in the working
+          # directory, so the YAML variant has to be named explicitly.
+          trivyignores: .trivyignore.yaml
 
       - name: Build scan summary
         id: scan-summary
--- a/.github/workflows/container-rescan.yaml
+++ b/.github/workflows/container-rescan.yaml
@@ -21,6 +21,14 @@ jobs:
       TRIVY_DISABLE_VEX_NOTICE: "true"
 
     steps:
+      # Only needed for .trivyignore.yaml: the scan target is the published
+      # image, not the working tree. The ignore file tracks the default branch,
+      # so an entry that expires takes effect on the next nightly run.
+      - name: Checkout
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+        with:
+          persist-credentials: false
+
       - name: Pull image
         env:
           IMAGE_REF: ${{ env.IMAGE_REF }}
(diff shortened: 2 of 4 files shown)


leoparente:
@unlisted this is required? Current issues does not block merge PRS

unlisted:
> @unlisted this is required? Current issues does not block merge PRS

Oh shoot. I misread the failed check as a blocker. No, it's not required. Maybe consider merging this or removing the check altogether because it is confusing. I will merge the release PR.

PR #529, opened 2026-08-12
fix(snmp-discovery): leave an IP unassigned when its interface was never walked

leoparente:
## Problem

Ref: ENGHLP-1566 (customer hit this on orb-agent 2.13.0 / snmp-discovery v1.36.0, i.e. latest).

An IP row whose `ipAddressIfIndex` (or legacy `ipAdEntIfIndex`) points at an ifIndex that **no interface PDU ever populated** causes `GetOrCreateEntity` to fabricate a placeholder `Interface` named `unknown` with no device and no type. That placeholder travels as the address's `assigned_object`. Diode then tries to auto-vivify a `dcim.interface` and NetBox rejects the entire bulk plan:

```
bulk plan error for entity 7765: {"dcim.interface":{"device":["Field device is required"],
                                  "type":["This field cannot be blank."]}}
```

Because IP-assigned interfaces are deliberately excluded from top-level emission, this nested stub is the *only* wire payload for that interface — nothing else supplies `device` or `type`.

On the device side this correlates with the interface table being only partially returned (an SNMP view/ACL exposing the IP tables but restricting `ifTable`, or a truncated walk).

## Why unassign rather than drop

The mapper cannot decide this at row-mapping time — whether the `ifTable` row eventually comes back is a function of walk order, so the existing `ipAddressIfIndex == 0` guard structurally cannot cover the non-zero-but-unwalked case. The guard therefore goes at finalization, where the registry's verified set is complete.

Given that, unassign beats drop:

- **Drop loses strictly more.** `DerivePrefixes` builds Prefixes *from* surviving IPAddress entities (`mapping/prefix.go:45-49`), so dropping `10.32.136.102/24` would also lose `10.32.136.0/24`, plus the whole IPAM payload (description, tags, tenant, role, VRF defaults) — silently.
- **An unassigned IP is already normal, shipped practice.** network-discovery emits *only* unassigned IPs — unreachably so, since its config exposes no device/interface knob (`network-discovery/policy/runner.go:363-365`).
- **It cannot clobber a good assignment.** An omitted `assigned_object` leaves the existing NetBox assignment intact: `protojson` runs without `EmitUnpopulated`, the plugin transformer only writes keys present in the incoming dict, and the applier uses DRF `partial=True`. Upstream regression test: `test_api_generate_diff.py:809`.
- **It self-heals.** The plugin matches `ipam.ipaddress` on host address + VRF, not `assigned_object` (`matcher.py:184-197`), so a later successful walk PATCHes the interface onto the same row — no duplicate.
- **In-backend precedent agrees**: `ifIndex=0` keeps the address unassigned, `dedupIPAddresses` treats a placeholder as unbound, `VlanMapper` skips placeholders. The two drop precedents are gated on operator intent (`interface_exclude_patterns`) or a deliberately-removed parent (stack member).

For contrast, device-discovery **cannot hit this at all**: NAPALM returns IP data keyed by interface *name*, so the reference and the identity are the same object and cannot dangle.

## Changes

1. **`dropUnverifiedInterfaceAssignments`** (`mapping/mapping.go`) — clears `AssignedObject` for any IPAddress bound to an unverified interface. Runs after `dedupIPAddresses` so cross-table legacy/modern selection still sees the original binding state. Safe for primary-IP selection, which already required a verified interface via `pickPrimaryIPHit`.

2. **VRF preservation** (`mapping/vrf.go`, `policy/runner.go`) — clearing the ref removes the only pointer `AttachVrfs` can key on, so the discovered VRF was being forfeited for exactly the addresses already degraded. That also risks a **duplicate NetBox row**, since IP identity is address+vrf and losing the VRF can route the address to the global matcher instead of the VRF-scoped one. The placeholder's ifIndex is now recorded before clearing, and `AttachVrfsToUnverified` reattaches from it. Purely additive: no existing signature or ordering changes, and a no-op unless `discover_vrfs` surfaced a VRF for that ifIndex. Attachments also land in `vrfByAddress` so the derived Prefix carries the same VRF.

3. **Hardened guard against a `{}` clear** — the wire contract distinguishes an **absent** `assigned_object_interface` ("leave untouched") from an **empty object** ("clear the FK"). The SDK gates the field on a raw `!= nil`, so a typed nil `(*diode.Interface)(nil)` passes that check and serializes to `{}`, silently unassigning a correct IP. `assert.Nil` cannot tell the two apart, so the guard uses a raw comparison and a new test asserts on the serialized proto-JSON.

Log level left at `Debug` per review.

## Validation

TDD throughout — every test was watched fail before the code existed.

**Unit** (`mapping/unverified_assignment_test.go`): dangling ifIndex leaves the address unassigned; a walked interface's assignment survives (no over-stripping); the cleared assignment's ifIndex is recorded; the VRF is reattached and recorded for prefix derivation; an unrelated ifIndex does not overwrite a configured VRF default; and the cleared assignment serializes with the key **absent**.

**Mutation-tested** — each new guard was proven to fail when its production code is broken:

| Mutation | Result |
|---|---|
| store a typed nil instead of untyped nil | ✅ fails — emits `{"address":…,"assignedObjectInterface":{}}` |
| drop the ifIndex recording | ✅ fails |
| make `AttachVrfsToUnverified` a no-op | ✅ fails |

That first row also **empirically confirms the `{}` clear hazard** is real rather than inferred.

**End to end** — snmpsim recording of a 3-member 2960X stack with one injected legacy `ipAddrTable` row bound to ifIndex 65000, which has no `ifTable`/`ifXTable` row (`c2960x-unknown-iface-ip.snmprec` in the test-lab repo). Ran the real binary in `-dry-run` against the sim:

| | `10.32.136.102/24` |
|---|---|
| before | `assigned_object_interface {name: "unknown"}`, `type` blank → NetBox rejects |
| after | `assigned_object_interface` **absent**, address ships clean |

Blast radius: entity count unchanged (156 → 156), **exactly one** entity differs, and the address is preserved rather than dropped. Control address `10.54.64.9/27` keeps its `Vlan99` / `virtual` assignment; `10.32.136.0/24` is still derived. The post-follow-up payload is **byte-identical** to the first validated run, confirming zero behaviour change on the default path where `discover_vrfs` is off.

Checks: `go test ./...` green for the whole snmp-discovery module (12/12), `gofmt` clean, `go vet` clean, `golangci-lint` 0 issues.

## Known limitation

The ifIndex itself is not preserved anywhere on the emitted entity (only a Debug log mentions it). Putting it in `IPAddress.Description` was considered and rejected — that would write agent-derived data into an operator-facing NetBox field.

### DIFF
--- a/orb-discovery/snmp-discovery/mapping/mapping.go
+++ b/orb-discovery/snmp-discovery/mapping/mapping.go
@@ -757,6 +757,7 @@ func (m *ObjectIDMapper) MapObjectIDsToEntity(objectIDs ObjectIDValueMap) []diod
 	// IP entirely.
 	m.dedupIPAddresses(uniqueEntities)
 	m.filterExcludedEntities(uniqueEntities)
+	m.dropUnverifiedInterfaceAssignments(uniqueEntities)
 
 	currentDevice := m.registry.GetOrCreateEntity(DeviceEntityType, CurrentDeviceIndex).(*diode.Device)
 
@@ -1285,6 +1286,37 @@ func (m *ObjectIDMapper) filterExcludedEntities(entities map[diode.Entity]bool) 
 			delete(entities, entity)
 			m.logger.Debug("excluding IP for excluded interface", "interface", *iface.Name)
 		}
+	}
+}
+
+// dropUnverifiedInterfaceAssignments clears an IPAddress's AssignedObject when
+// the interface it points at was never populated by an InterfaceMapper.Map call
+// during the walk. This happens when ipAddressIfIndex (or the legacy
+// ipAdEntIfIndex) references an ifIndex whose ifTable/ifXTable row never came
+// back: GetOrCreateEntity fabricates a placeholder Interface named
+// DefaultInterfaceName with no device and no type. Emitting that placeholder
+// makes Diode auto-vivify a dcim.interface which NetBox rejects (device
+// required, type blank), failing the whole bulk-plan. Leaving the address
+// unassigned mirrors the ipAddressIfIndex=0 handling in the assignedObject
+// mapper and is safe for primary-IP selection, which already requires a
+// verified interface via pickPrimaryIPHit.
+//
+// Runs after dedupIPAddresses so cross-table dedup still sees the original
+// binding state when choosing between the legacy and modern rows.
+func (m *ObjectIDMapper) dropUnverifiedInterfaceAssignments(entities map[diode.Entity]bool) {
+	for entity := range entities {
+		ip, ok := entity.(*diode.IPAddress)
+		if !ok || ip.AssignedObject == nil {
+			continue
+		}
+		if m.registry.hasVerifiedInterface(ip) {
+			continue
+		}
+		if iface, ok := ip.AssignedObject.(*diode.Interface); ok && iface.Name != nil {
+			m.logger.Debug("dropping IP assignment to unwalked interface; leaving address unassigned",
+				"address", derefString(ip.Address), "interface", *iface.Name)
+		}
+		ip.AssignedObject = nil
 	}
 }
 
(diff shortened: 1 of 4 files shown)


leoparente:
@codex review

samiura:
LGTM, ignoring the vulnerability failures!

samiura (approved):

leoparente:
@codex review

PR #544, opened 2026-08-19
fix(ci): check out the ignore file before the scheduled rescan

leoparente:
## What broke

The nightly `Scheduled Container Rescan` failed on its first run after
[#542](https://github.com/netboxlabs/orb-agent/pull/542) landed
([run 32222170461](https://github.com/netboxlabs/orb-agent/actions/runs/32222170461/job/95974556637)):

```
ERROR: cannot find ignorefile '.trivyignore.yaml'.
##[error]Process completed with exit code 1.
FATAL  Fatal error  file open error: open trivy-output.json: no such file or directory
##[error]Path does not exist: trivy-results.sarif
```

One missing file produced three failing steps: the scan aborted before writing
`trivy-output.json`, then `trivy convert` and the SARIF upload each failed on the
artifact that was never created.

It had passed every night up to 2026-08-18 and broke on the first run against
`21cd9ed`, so the cause is unambiguous.

## Why

`#542` wired `trivyignores: .trivyignore.yaml` into both scanning workflows, but
the two workflows are not alike:

| | `build.yaml` | `container-rescan.yaml` |
|---|---|---|
| checks out the repo | yes, it builds the image from source | **no**, it pulls a published image |
| so the ignore file is | on disk | absent |

Nothing in the rescan job read the working tree before, so it had no reason to
check anything out. Adding a file reference to the scan gave it one.

## The fix

A sparse checkout of that single file, rather than the whole tree. The scan
target is a pulled image and nothing else in the job reads the repo, so the
checkout exists only to place the ignore file.

Dropping the `trivyignores` input would be the wrong repair. This job uploads its
results to code scanning, so without the ignore file the two accepted
vendored-pip findings would come back as alerts, which is exactly what the file
exists to prevent.

## Verification

Not taken on trust a second time, since the first attempt at this wiring is what
broke. Details in the comment below.

### DIFF
--- a/.github/workflows/container-rescan.yaml
+++ b/.github/workflows/container-rescan.yaml
@@ -21,6 +21,15 @@ jobs:
       TRIVY_DISABLE_VEX_NOTICE: "true"
 
     steps:
+      # The scan needs .trivyignore.yaml from the repo. Nothing else here does,
+      # since the image is pulled rather than built.
+      - name: Check out the ignore file
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+        with:
+          sparse-checkout: .trivyignore.yaml
+          sparse-checkout-cone-mode: false
+          persist-credentials: false
+
       - name: Pull image
         env:
           IMAGE_REF: ${{ env.IMAGE_REF }}


jajeffries (approved):

leoparente:
## Verification

### 1. The workflow now runs, proven by running it

Dispatched this workflow from this branch:
[run 32249819450](https://github.com/netboxlabs/orb-agent/actions/runs/32249819450) → **success**.

```
Check out the ignore file    Setting up sparse checkout
                            git config core.sparseCheckout true
Scan for vulnerabilities    Using YAML ignorefile '.trivyignore.yaml':
Upload SARIF                (ok)
```

`Using YAML ignorefile` replaces the previous `cannot find ignorefile`, and all
three previously-failing steps completed: scan, convert, SARIF upload.

### 2. The sparse pattern really does place that file

Checked the pattern independently, since a sparse checkout that silently matched
nothing would reproduce the original failure:

```
$ git sparse-checkout set --no-cone .trivyignore.yaml && git checkout origin/develop
$ ls -a
.trivyignore.yaml
```

One file, which is the whole intent.

### 3. The acceptance holds against the real published image

The earlier verification used a minimal reproducer image. This checks the actual
artifact this job scans, `netboxlabs/orb-agent:develop`:

| | findings | blocking |
|---|---|---|
| no ignore file | 3 (`GHSA-6v7p-g79w-8964`, `CVE-2025-47273`, `CVE-2026-59890`) | HIGH present |
| with the ignore file | 1 (`CVE-2026-59890`) | none |

Two things worth noting from that first row. The 32 Go stdlib CVEs are gone from
the real image, so #542's toolchain fix landed as intended. And the three
remaining findings are exactly the three predicted, with nothing unexpected.

### 4. The ignore file's factual claims, checked on the shipped image

Both statements are load-bearing for the acceptance, so both were verified
against the real image rather than inferred:

```
$ docker run --rm --entrypoint sh netboxlabs/orb-agent:develop \
    -c 'ls /usr/local/lib/python3.14/site-packages/pip/_vendor/'
... cachecontrol certifi distlib distro idna msgpack packaging pkg_resources ...
$ pip --version
pip 26.2.1
```

`setuptools` is absent from `_vendor/` (only `pkg_resources` ships), which is
what makes that entry metadata-only. `msgpack` is present, which is why that
entry is written as a narrower argument rather than the same one.

PR #546, opened 2026-08-19
feat(agent): add GitHub App auth to git config manager

cflanderscpc:
## Summary

Adds `auth: github_app` as a new authentication mode for `config_manager.sources.git`, alongside the existing `basic` and `ssh` modes. Instead of using a personal identity with a static PAT or SSH key sitting on disk indefinitely, the agent authenticates to GitHub as a GitHub App Installation and mints a short-lived (~1 hour), narrowly-scoped installation token on every policy-fetch cycle.

This is primarily a security hardening change for anyone running the Git config manager against GitHub Enterprise Cloud as it removes the need to distribute and rotate a long-lived credential to every agent.

## Motivation

The Git config manager currently supports two credentialed auth modes, and both rely on a static secret that lives for as long as nobody rotates it:

`auth: basic` — a username + password/PAT, passed straight through as HTTP Basic auth
`auth: ssh` — a private key file (+ optional passphrase) mounted on the agent host

Both work, but they share the same underlying problem: the credential doesn't expire on its own, it's usually broader in scope than "read this one policy repo," and if it leaks due to a compromised host, misconfigured volume mount, log capture, etc., it stays valid until a human notices and manually revokes it. PATs are also typically tied to a person's account rather than the agent itself, which muddies audit trails and complicates offboarding.

GitHub Apps address this directly:
- Replaces a long-lived Git credential with short-lived installation access tokens. The GitHub App private key remains a long-lived credential and must be protected and rotated, but it provides centralized revocation and separates Git activity from individual user accounts.
- Least privilege: The App can be created/installed with only `contents:read` access to the specific policy repo(s) it needs and nothing else
- Clean audit trail: Git operations show up as the App/Installation rather than an individual's PAT
- Centralized revocation: Disabling/Uninstalling the app or rotating its key once is all that's necessary rather than hunting down copies of a PAT across every agent host
- Policy-aligned: A growing number of organizations restrict or block classic PATs outright and put extra scrutiny towards fine-grained PATs. `github_app` provides a way for those organizations to continue using the Git config manager instead of `basic` auth breaking.

### DIFF
--- a/agent/config/types.go
+++ b/agent/config/types.go
@@ -25,11 +25,11 @@ type LocalManager struct {
 // ConfigManager. Only github.com is supported (including GitHub Enterprise
 // Cloud, which is served from github.com / api.github.com).
 type GitHubAppAuth struct {
-	// AppID is the JWT issuer. Prefer the app's Client ID (Iv23li...), which is
+	// ClientID is the JWT issuer. Prefer the app's Client ID (Iv23li...), which is
 	// what the GitHub API documentation now specifies; the numeric App ID from
 	// the app's settings page is still accepted, but is the legacy form and may
 	// be dropped in a future API version.
-	AppID string `yaml:"app_id"`
+	ClientID string `yaml:"client_id"`
 	// InstallationID identifies the installation of the app on the account that
 	// owns the repository. It is the numeric id in the URL of the installation's
 	// settings page - not the App ID.
--- a/agent/config/types.go
+++ b/agent/config/types.go
@@ -21,16 +21,35 @@ type LocalManager struct {
 	Config string `yaml:"config"`
 }
 
+// GitHubAppAuth holds GitHub App installation credentials for the Git
+// ConfigManager. Only github.com is supported (including GitHub Enterprise
+// Cloud, which is served from github.com / api.github.com).
+type GitHubAppAuth struct {
+	// AppID is the JWT issuer. Prefer the app's Client ID (Iv23li...), which is
+	// what the GitHub API documentation now specifies; the numeric App ID from
+	// the app's settings page is still accepted, but is the legacy form and may
+	// be dropped in a future API version.
+	AppID string `yaml:"app_id"`
+	// InstallationID identifies the installation of the app on the account that
+	// owns the repository. It is the numeric id in the URL of the installation's
+	// settings page - not the App ID.
+	InstallationID string `yaml:"installation_id"`
+	// PrivateKey is the path to the app's .pem private key, or the PEM document
+	// itself when injected by ${VAR} or a secrets manager.
+	PrivateKey string `yaml:"private_key"`
+}
+
 // GitManager represents the Git ConfigManager configuration.
 type GitManager struct {
-	URL        string  `yaml:"url"`
-	Branch     string  `yaml:"branch"`
-	Auth       string  `yaml:"auth"`
-	Schedule   *string `yaml:"schedule,omitempty"`
-	Username   string  `yaml:"username"`
-	Password   string  `yaml:"password"`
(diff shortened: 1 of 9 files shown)


leoparente:
@codex review

agent/config/types.go
```diff
@@ -21,16 +21,35 @@ type LocalManager struct {
...
 
+// GitHubAppAuth holds GitHub App installation credentials for the Git
+// ConfigManager. Only github.com is supported (including GitHub Enterprise
+// Cloud, which is served from github.com / api.github.com).
+type GitHubAppAuth struct {
+	// AppID is the JWT issuer. Prefer the app's Client ID (Iv23li...), which is
+	// what the GitHub API documentation now specifies; the numeric App ID from
+	// the app's settings page is still accepted, but is the legacy form and may
+	// be dropped in a future API version.
+	AppID string `yaml:"app_id"`
```
leoparente: if we prefer client_ID and github calls it client_id in the examples and its code https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/generating-a-json-web-token-jwt-for-a-github-app?apiVersion=2022-11-28&productId=rest&utm_source=chatgpt.com#example-using-python-to-generate-a-jwt

we should rename it to `client_id` then in every place and docs

agent/configmgr/git.go
```diff
@@ -380,6 +394,20 @@ func (gc *gitConfigManager) Start(_ context.Context, cfg config.Config, backends
 		} else {
 			gc.authMethod, err = ssh.NewSSHAgentAuth("git")
 		}
+	case "github_app":
```
leoparente: Maybe it is good to add auth type validation to avoid typos, but keep in mind that None is valid. something like:

```
type GitAuthMode string

const (
    GitAuthNone      GitAuthMode = ""
    GitAuthBasic     GitAuthMode = "basic"
    GitAuthSSH       GitAuthMode = "ssh"
    GitAuthGitHubApp GitAuthMode = "github_app"
)

switch GitAuthMode(gc.config.Auth) {
case GitAuthNone:
case GitAuthBasic:
    // ...
case GitAuthSSH:
    // ...
case GitAuthGitHubApp:
    // ...
default:
    return fmt.Errorf(
        "unsupported git auth mode %q; expected basic, ssh, github_app, or empty",
        gc.config.Auth,
    )
}
```

agent/config/types.go
cflanderscpc: renamed to all app id instances to client id

agent/configmgr/git.go
cflanderscpc: Added the typo guard :)

leoparente:
@codex review

leoparente:
@codex review

agent/configmgr/git_github_app.go
```diff
@@ -0,0 +1,481 @@
...
+			"(scp-style and ssh URLs cannot use installation tokens)", rawURL)
+	}
+	if !strings.EqualFold(u.Scheme, "https") {
+		return fmt.Errorf("github_app: url uses scheme %q; github_app auth requires https", u.Scheme)
+	}
+	if host := strings.ToLower(u.Hostname()); host != "github.com" {
+		return fmt.Errorf("github_app: url host %q is not github.com; github_app auth supports github.com "+
+			"only (for GitHub Enterprise Server use auth: basic with a personal access token)", host)
+	}
+	if host := strings.ToLower(u.Hostname()); host == "www.github.com" {
```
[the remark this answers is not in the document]
cflanderscpc: Already fixed in the subsequent commit

leoparente:
@codex review

leoparente (approved):
LGTM! Thanks again for the contribution!

Issue #537, opened 2026-08-14
fortinet_get_system_interface + fortinet_get_system_interface_physical templates fail to parse FortiOS 7.4.12 output


jmorawc:
## Environment
- FortiOS version: 7.4.12
- Platform: `fortinet`
- Used via: `fortinet_fortios_ssh` NAPALM driver, which calls `ntc_templates.parse_output()` for both commands during interface discovery

## How I found this

I was debugging why a NAPALM-based discovery tool (`fortinet_fortios_ssh` driver) wasn't returning any interfaces for a FortiGate-100F running 7.4.12. Both getters catch parsing exceptions broadly and return an empty dict, so the failure was silent until we reproduced it manually with `ntc_templates.parse_output()` directly against the device's raw CLI output.

---

### Command
`get system interface physical`

### Error
Running the raw output from a 7.4.12 device through the template raised:
`TextFSMError: State Error raised. Rule Line: 39. Input Line: medium: n/a`

### Root cause
FortiOS 7.4.12 added a new `medium:` line to the per-interface status block (likely indicating copper/fiber media type). The `PhysicalStatus` state in the template has no rule matching this line, so it falls through to `^. -> Error` and the whole parse aborts.

### Temp Fix
Adding 
`Value MEDIUM (\S+)`
and
`^\s+medium:\s+${MEDIUM}$$`
into .textfsm file

After adding these lines, the same raw output parsed cleanly with no errors.

---

### Command
`get system interface`

### Error:
TextFSMError: State Error raised. Rule Line: 62. Input Line: name: [interface name] mode: static ip: 0.0.0.0 0.0.0.0 status: up netbios-forward: disable type: physical netflow-sampler: disable sflow-sampler: disable src-check: enable explicit-web-proxy: disable explicit-ftp-proxy: disable proxy-captive-portal: disable trunk: disable wccp: disable drop-overlapped-fragment: disable drop-fragment: disable mtu-override: disable

### Root Cause:

FortiOS 7.4.12 added several new fields that none of the template's 8 hardcoded line variants account for:
- `trunk:` — present on nearly every interface
- `switch:` — present on VLAN interfaces bound to a switch fabric
- `aggregate:` — present on physical ports that are members of a link aggregation group

### My Temp Fix:

Changing content of fortinet_get_system_interface.textfsm

to:
```
Value Required NAME (.+?)
Value MODE (\S+)
[...]

PR #547, opened 2026-08-21
fix(device-discovery): parse FortiOS interfaces without ntc-templates

leoparente:
Fixes the parse failure reported in netboxlabs/orb-agent#537. Deliberately not using a closing keyword:<br>the issue stays open until the reporter confirms on their 7.4.12 device, and there is<br>an open request there for sanitised captures to replace the reconstructed fixtures.

## The bug

FortiOS 7.4.12 added a `medium:` line to `get system interface physical`, and<br>`trunk:` / `switch:` / `aggregate:` to `get system interface`. Neither upstream<br>ntc-template admits them, both end in `^. -> Error`, and every getter caught the<br>exception, logged at DEBUG and returned `{}`. So a FortiGate-100F reported **no****<br>****interfaces at all**, and nothing said why — the reporter had to reproduce the parse by<br>hand to find it.

The two templates enumerate every field on the line. The driver reads **six**:

| command | fields used |
| -- | -- |
| `get system interface physical` | `name`, `status`, `speed` |
| `get system interface` | `name`, `ip_address`, `netmask` |

That mismatch is the defect: the parse is coupled to fields nobody reads, so any<br>FortiOS release adding one breaks discovery.

## The fix

Parse both commands in the driver, tolerantly. Unknown fields are ignored.

This extends a pattern this file already established — `get_facts` abandoned the<br>template for `get system status` with the comment *"has* `^. -> Error` *and does not**<br>**cover all lines real FortiOS devices emit"*.

Same input, before and after:

```
get system interface physical
  ntc-template : TextFSMError: State Error raised. Rule Line: 39
  local parser : 2 rows, 0 anomalies -> ['port1', 'port2']

get system interface
  ntc-template : TextFSMError: State Error raised. Rule Line: 62
  local parser : 2 rows, 0 anomalies -> ['port1', 'port2']
```

## Why not fix the templates

Three alternatives were tried and measured, not waved away.

**Patch the templates, in-repo or upstream.** A tolerant TextFSM template *is*<br>writable — that objection was wrong. What kills it is mechanism: this package's build<br>silently drops `.textfsm` files from the wheel (`[tool.setuptools.packages.find]`, no<br>`package-data`, no `MANIFEST.in`), and `parse_output(try_fallback=True)` cannot catch<br>this failure class at all, because `TextFSMError` is not a `clitable.CliTableError`.<br>An upstream PR is still worth sending, but releases are quarterly, so it cannot be the<br>fix here.

**Try the template first, fall back on exception.** The template is not the integrity<br>check that depends on: two consecutive `==[name]` headers parse *successfully* and<br>silently drop the first interface. The fallback would also get no CI coverage, since it<br>only runs on inputs where the template raises.

**One shared tokenizer for every FortiOS command.** Unsafe: in `get system status` the<br>keys are multi-word English, so `CPU usage:` and `Memory usage:` both reduce to<br>`usage`. The scanner is shared between the two interface commands only.

## The parity baseline

`tests/.../corpus/` vendors the 11 real FortiOS captures from<br>[networktocode/ntc-templates](<https://github.com/networktocode/ntc-templates>)<br>`tests/fortinet/` (Apache-2.0) — six release trains, **73 physical blocks and 102 flat****<br>****rows**. Routable addresses are replaced with RFC 5737 space; verified against fresh<br>upstream copies that line counts match and every differing line differs only in an<br>address token, so indentation, field order and the 7.0 capture's CRLF endings are<br>provably intact.

`baseline.json` records what the templates return for the six consumed fields, and<br>`test_corpus_parity.py` holds the new parsers to it: **175/175 rows match, zero****<br>****anomalies**. That test is the only thing that mechanically catches a renamed row key or<br>a block dropped at end of input, so it landed before the parsers.

## Observability

The failure mode changes shape: a new field is now ignored on purpose. "Warn when<br>nothing parsed" is necessary but not sufficient, because every defect these parsers<br>guard against yields a **non-zero** row count. Three signals:

1. non-empty output, zero rows — the state the reporter hit
2. anomalies counted — an unreadable header, a truncated block, a device answering with<br>an error string
3. rows carried an address-shaped value yet none was emitted — catches a future rename<br>of `ip:` silently emptying every address in NetBox

Signal 3 is keyed on the value's **shape**, not the `ip_address` name. Keyed on the name<br>it is dead code, because both failures it exists to catch leave `ip_address` absent.<br>Management keys are excluded, or a box whose interfaces are all unnumbered but which<br>reports a management address warns on every poll.

Unknown field keys log once per distinct key per parse at DEBUG — the drift sensor that<br>dropping `^. -> Error` gives up, and what tells us to send a patch upstream.

## Testing

`2307 passed, 161 skipped` (2273 on develop), ruff clean. Every guard is<br>mutation-verified rather than assumed:

| mutation | result |
| -- | -- |
| drop the end-of-input close | `modem` vanishes from the first capture |
| drop `ring-rx` from the allowlist | fails, naming the missing key |
| remove physical signal 1 / flat signal 1 / neuter signal 3 | each fails its own test |
| treat an unknown-field line as unreadable | fails 4 tests, including the 7.4.12 scenarios |
| relax `status` **and** `speed` to `or` | fails the two boundary tests |

Three reconstructed 7.4.12 scenarios, each labelled as reconstructed in its README,<br>each built only from fragments the issue actually shows:

* `test_get_interfaces/7412` — `medium: n/a` placed deliberately *above* `status:`.<br>Below it, the scenario passes either way; above it, a regression that treats an<br>unknown field line as unreadable discards the block and the scenario fails.
* `test_get_interfaces_ip/7412` — carries an invented second addressed interface,<br>because with only the unnumbered port1 the expected result is `{}` and the fixture<br>could be **truncated to zero bytes** and still pass.
* `test_get_facts/7412` — the only test that pins that getter's rewiring. Without<br>it, `get_facts` can be left calling the aborting template with the whole suite green.

## Notes for review

* **All three getters keep their** `try/except → return {}` **guard**, with the log raised<br>to WARNING. `policy/runner.py` calls these three inline with no handler of its own,<br>unlike every other getter it invokes, so an escaping exception costs the whole device<br>its entities. The silence was the bug, not the guard.
* **The committed flat fixture was not device-shaped** — one `== [ root ]` header above<br>two `name:` lines, where all seven real captures are strictly 1:1 with the interface<br>name in the bracket. Replaced. Nothing is derived from the header, not even as a<br>fallback: one capture has `== [ VPN-TUN ]` above `name: VPN-LAB`.
* **No lint suppressions.** `_parse_physical` initially carried a `# noqa: C901` for<br>complexity 11; it is now split into `_close_physical_block` and<br>`_absorb_physical_fields` at its two real seams, so the file has no `noqa` at all.<br>The guards were re-mutation-checked after the restructure.
* **Deliberately dropped** relative to the templates: `duplex`, `ipv6_address`,<br>`ipv6netmask`, and the physical `ip` left unsplit. All 73 real blocks report<br>`ipv6: ::/0`, and the getter that would use an address reads the flat command, which<br>emits no `ipv6:` at all.
* `is_enabled` **is unconditionally true on FortiOS** — all 175 observed status values<br>are `up` or `down`; `disabled` never appears. Recorded in the docstring as deliberate.

## Follow-ups, not in this PR

* Ask the reporter for sanitised full output, and specifically for lines containing<br>`switch:` and `aggregate:`, whose positions and value shapes are currently invented.
* Send an upstream ntc-templates PR. **Not the patch in the issue**: measured, it fails<br>on 7 of 7 real captures (97 of 102 lines) for four independent reasons — it reorders<br>`mtu-override:`, makes `status:` mandatory while four lines omit it, has no group for<br>`ring-rx:`/`ring-tx:`, and ends every optional group in `\s+`. Fixing only the<br>whitespace still fails all seven.
* The \~40 other `except Exception → logger.debug → return {}` sites across the custom<br>drivers are the same bug class, and want their own change.


leoparente:
@codex review

leoparente:
@codex review

mfiedorowicz (approved):
LGTM

Issue #514, opened 2026-08-07
discover_modules: full does not discover transceivers on fixed-port C9500-16X StackWise Virtual


oyabiccha:
## Environment

- Orb Agent: 2.12.0
- Device Discovery backend: 1.31.0
- NetBox: 4.5.8
- Device: Cisco Catalyst C9500-16X
- Topology: 2-member StackWise Virtual
- Driver: ios

## Configuration

Module discovery is enabled with:

    config:
      discover_modules: full
      defaults:
        site: <site-name>

The device itself is discovered successfully, and the StackWise Virtual pair is correctly represented in NetBox as a Virtual Chassis with two member devices.

## Expected behavior

Transceivers reported by `show inventory` should be discovered as NetBox Module / ModuleBay objects when `discover_modules: full` is enabled.

For example, the device reports the following transceivers:

    NAME: "Te1/0/1", DESCR: "SFP-10GBase-LR"
    PID: SFP-10G-LR, VID: V02, SN: REDACTED

    NAME: "Te1/0/3", DESCR: "1000BaseLX SFP"
    PID: GLC-LH-SMD, VID: V01, SN: REDACTED

    NAME: "Te2/0/14", DESCR: "SFP-10GBase-SR"
    PID: SFP-10G-SR, VID: V03, SN: REDACTED

I would expect these optics to appear as modules associated with their respective interfaces.

## Actual behavior

No Module or ModuleBay objects are created.

The number of ingested entities is exactly the same with and without module discovery enabled.

Without `discover_modules: full`:

    Successfully ingested 332 entities

With `discover_modules: full`:

    Successfully ingested 332 entities

There are no warnings or errors related to module discovery in the Orb Agent output.

## Relevant show inventory

The relevant inventory structure is:

    NAME: "c95xx Stack", DESCR: "c95xx Stack"
    PID: C9500-16X, VID: V01, SN: REDACTED

    NAME: "Switch 1", DESCR: "C9500-16X"
    PID: C9500-16X, VID: V01, SN: REDACTED

    NAME: "Te1/0/1", DESCR: "SFP-10GBase-LR"
    PID: SFP-10G-LR, VID: V02, SN: REDACTED

    NAME: "Te1/0/2", DESCR: "SFP-10GBase-LR"
    PID: SFP-10G-LR, VID: V02, SN: REDACTED

    NAME: "Te1/0/3", DESCR: "1000BaseLX SFP"
[...]

PR #548, opened 2026-08-24
fix(device-discovery): warn on unrecognized config and options keys

leoparente:
## Why

A policy key written at the wrong nesting level parses without complaint and is then
dropped, so the setting never takes effect and nothing says so.

This is how #514 stayed open for two weeks. The reporter had:

```yaml
config:
  discover_modules: full   # belongs under config.options
  defaults:
    site: <site>
```

`config.options` came out `None`, the runner gate never opened, `get_modules()` was never
called. Every symptom pointed at the hardware instead: entity count identical with and
without the option, no warnings anywhere, nothing in the reconciler logs, and the same
result on standalone, stack and StackWise Virtual. A fix shipped for that report was never
once exercised on the reporter's device.

A completely invented key was accepted just as silently.

## What

Keeps the permissive parse and adds a warning naming the key, plus the path where it would
have been recognized when that is within two levels:

```
WARNING device_discovery.policy.unknown_keys: ignoring unrecognized config key
'discover_modules'; did you mean 'options.discover_modules'?
```

Scoped deliberately to `config` and `options`, whose keys are a closed documented set. The
policy map, `scope` entries and the surrounding YAML carry operator-chosen keys and
anchors with no set to check against, so warning there would fire on correct files and
train operators to ignore the warning.

Not `extra="forbid"`: that would turn any existing config with a stray key into a hard
failure on upgrade.

## Verification

- 2320 passed, 161 skipped. The only warnings this fires across the whole suite are the
  four from its own tests, so it does not fire on any existing valid config.
- Confirmed end to end through the real `PolicyManager` with the reporter's exact policy.
- `ruff check` and `ruff format --check` clean.

## Notes

The other three backends have the same silent-ignore behaviour (`snmp-discovery`,
`gnmi-discovery`, `network-discovery` all use plain `yaml.Unmarshal`). Verified by test:
both a misplaced `discover_modules` and a `totally_bogus_key` parse without error. Those
follow in their own PRs, one per scope.

### DIFF
--- a/orb-discovery/device-discovery/device_discovery/policy/models.py
+++ b/orb-discovery/device-discovery/device_discovery/policy/models.py
@@ -12,6 +12,7 @@ from typing import Any, Literal
 from croniter import CroniterBadCronError, croniter
 from pydantic import BaseModel, Field, field_validator
 
+from device_discovery.policy.unknown_keys import WarnUnknownKeys
 from device_discovery.stack_naming import (
     DEFAULT_STACK_MEMBER_TEMPLATE,
     stack_template_problem,
@@ -237,7 +238,7 @@ class Defaults(BaseModel):
         return v
 
 
-class Options(BaseModel):
+class Options(WarnUnknownKeys):
     """Model for discovery options."""
 
     platform_omit_version: bool | None = Field(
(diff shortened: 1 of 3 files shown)


leoparente:
@codex review

samiura (approved):

PR #551, opened 2026-08-24
fix(network-discovery): warn on unrecognized config keys

leoparente:
## Why

Last of four, same class of bug as #548, #549 and #550. `yaml.Unmarshal` ignores
unrecognized keys, so a policy key written at the wrong nesting level decodes without
complaint and is then dropped. The setting never takes effect and nothing says so.

## What

A second decode with `KnownFields(true)` purely to harvest the report, logged as warnings.
The error is never returned, so an unrecognized key stays non-fatal.

This backend is the odd one of the four: it has no `options` block, and almost all of its
tuning lives on `scope` — `ports`, `fast_mode`, `timing`, `scan_types`, `top_ports`. So the
misplacement that actually bites here is a scope key written under `config`, and the
warning names the block it belongs to:

```
msg="ignoring unrecognized config key" key=fast_mode did_you_mean=scope.fast_mode
```

Defaults keys get the same treatment (`defaults.network_mask`). Both paths come from
reflection over this backend's own structs, so they cannot drift from the real field sets.

The report is filtered to `config.PolicyConfig`. The policy map, scope entries and
surrounding YAML anchors are excluded: warning there would fire on correct files and train
operators to ignore the warning.

## Verification

- 5 packages pass. 7 tests cover a scope key on config, a defaults key on config, an
  invented key with no suggestion, silence outside the config block, silence on a valid
  policy, a syntax error left to the real decode, and a nil logger.
- No existing policy fixture triggers the warning.
- `make lint-all` clean for all Go modules.

Completes the set: #548 device-discovery, #549 snmp-discovery, #550 gnmi-discovery.

### DIFF
--- a/orb-discovery/network-discovery/config/unknown_keys.go
+++ b/orb-discovery/network-discovery/config/unknown_keys.go
@@ -1,0 +1,104 @@
+package config
+
+import (
+	"bytes"
+	"errors"
+	"log/slog"
+	"reflect"
+	"regexp"
+	"strings"
+
+	"gopkg.in/yaml.v3"
+)
+
+// yaml.v3 ignores unrecognized keys, so a key written at the wrong nesting
+// level decodes without complaint and is then dropped: the setting silently
+// never takes effect, and nothing downstream can tell that apart from the
+// option being absent. WarnUnknownPolicyKeys keeps the permissive decode and
+// reports those keys instead.
+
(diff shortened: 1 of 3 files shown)


leoparente:
@codex review

samiura (approved):

leoparente:
@codex review

Issue #554, opened 2026-08-25
ssh key auth no longer working


lgwapnitsky:
No updates to the orb agent, but it is no longer recognizing my ssh key for when I do a git clone.

```
{"time":"2026-08-25T18:26:12.68656204Z","level":"INFO","msg":"detected default branch","branch":"main"}
{"time":"2026-08-25T18:26:12.686582499Z","level":"INFO","msg":"cloning repository","url":"ssh://git@umitgit:222/diode/orb_launch","branch":"main"}
{"time":"2026-08-25T18:26:12.839511911Z","level":"ERROR","msg":"agent startup error","error":"yaml: line 89: did not find
expected key"}
```

Yet, if I do a test to the git server, I receive a successful response with the identical key:

```
➜ ssh -vT git@umitgit -p 222 -i /opt/orb/id_ed25519
debug1: OpenSSH_10.0p2 Debian-7+deb13u4, OpenSSL 3.5.6 7 Apr 2026
debug1: Reading configuration data /etc/ssh/ssh_config
debug1: Reading configuration data /etc/ssh/ssh_config.d/20-systemd-ssh-proxy.conf
debug1: /etc/ssh/ssh_config line 21: Applying options for *
debug1: Connecting to umitgit [10.3.1.111] port 222.
debug1: Connection established.
debug1: identity file /opt/orb/id_ed25519 type 3
debug1: identity file /opt/orb/id_ed25519-cert type -1
debug1: Local version string SSH-2.0-OpenSSH_10.0p2 Debian-7+deb13u4
debug1: Remote protocol version 2.0, remote software version OpenSSH_10.2
debug1: compat_banner: match: OpenSSH_10.2 pat OpenSSH* compat 0x04000000
debug1: Authenticating to umitgit:222 as 'git'
debug1: load_hostkeys: fopen /root/.ssh/known_hosts2: No such file or directory
debug1: load_hostkeys: fopen /etc/ssh/ssh_known_hosts: No such file or directory
debug1: load_hostkeys: fopen /etc/ssh/ssh_known_hosts2: No such file or directory
debug1: SSH2_MSG_KEXINIT sent
debug1: SSH2_MSG_KEXINIT received
debug1: kex: algorithm: mlkem768x25519-sha256
debug1: kex: host key algorithm: ssh-ed25519
debug1: kex: server->client cipher: chacha20-poly1305@openssh.com MAC: <implicit> compression: none
[...]

PR #555, opened 2026-08-25
fix(agent): name the file in git config parse errors

leoparente:
## Problem

Reported in #554. A config repository with a malformed YAML file failed startup with the bare parser error:

```
{"level":"INFO","msg":"detected default branch","branch":"main"}
{"level":"INFO","msg":"cloning repository","url":"ssh://git@umitgit:222/diode/orb_launch","branch":"main"}
{"level":"ERROR","msg":"agent startup error","error":"yaml: line 89: did not find expected key"}
```

The message names neither the file nor the step. Sitting directly under `cloning repository`, it reads as a clone failure, and the reporter filed it as SSH key authentication breaking. Authentication had in fact succeeded three times by that point: the fetch, the ref listing that produced `detected default branch`, and the clone itself.

A selector can point at many policy files, so on a real repository the error gives an operator nothing to act on.

## Change

`processSelector` and `readPolicies` now attribute their read and parse failures to the file:

```
failed to parse policies/broken.yaml: yaml: line 89: did not find expected key
```

This is the pattern `decodeFile` already uses for the local config file, so the two paths now report the same way.

## Tests

Two tests build a repository containing a malformed file and assert the startup error names it, one for `selector.yaml` and one for a selected policy file. Both fail on `develop` with `"yaml: line 1: did not find expected key" does not contain "selector.yaml"`.

`go test -race ./agent/configmgr/` and the full `./agent/...` suite pass; `make lint` reports 0 issues.

### DIFF
--- a/agent/configmgr/git.go
+++ b/agent/configmgr/git.go
@@ -94,7 +94,7 @@ func (gc *gitConfigManager) readPolicies(tree *object.Tree, matchingPolicies []s
 
 		reader, err := file.Reader()
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("failed to read %s: %w", path, err)
 		}
 		defer func() {
 			if err := reader.Close(); err != nil {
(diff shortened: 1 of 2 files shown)


leoparente:
@codex review

jajeffries (approved):

PR #557, opened 2026-08-26
fix(device-discovery): promote IOS fixed uplinks numbered off the baseboard

leoparente:
address #556.

### What

`discover_modules: full` emitted **nothing at all** on a C9200L-24PXG-4X. No modules, no errors, and an entity count identical with the option on or off.

### Why

The chassis has **fixed** 10G uplinks, but Cisco numbers them `Te1/1/x` — module 1, not the baseboard 0 — and ships no FRU row for them, because there is no removable module to report. `_optic_parent_is_baseboard` requires the module to be 0 in switch-prefixed mode, so every optic was declined, `bays_by_member` stayed empty, and `get_modules()` returned `None`.

Reproduced deterministically against `develop` with the reporter's exact `show inventory` rows:

```
distinct_switch_count=1  vc_mode=False  switch_prefixed=True
after parse:               bays_by_member={}   transceivers=['Te1/1/1','Te1/1/2']
                           claimed_slots=set()
declining promotion of TenGigabitEthernet1/1/1 onto member None slot 1
declining promotion of TenGigabitEthernet1/1/2 onto member None slot 1
PAYLOAD = None
```

### Why an allowlist rather than a slot rule

Nothing in the ifname distinguishes a fixed-uplink chassis that reports no card row from a modular chassis whose card row the vendor omitted. Both are `Te<member>/1/<port>` with no FRU row and no claimed slot. The chassis PID is the only available signal, so recognition is an explicit allowlist — the same approach `aruba_aoscx` (`_ARUBA_FIXED_FAMILIES`) and `nokia_sros` already take for this problem.

### Safety properties

The gate sits **after** the claimed-slot and parent-bay checks, so reported evidence always wins. `tests/custom_drivers/cisco_ios/test_optic_promotion_precedence.py` pins the three rules directly against `_attach_transceivers`:

| rule | outcome |
|---|---|
| bay built from a real card row | optic nests under it, no promotion |
| slot claimed by an unusable row | optic declined |
| no parent, no claim | promotes only when the chassis PID is allowlisted |

Two of those states cannot occur on real hardware — a fixed-uplink chassis never reports a removable uplink module. They are asserted at the helper level **on purpose**, as synthetic invariants, rather than as a driver fixture that would assert such a device exists.

Each of the three was verified by mutation: breaking the corresponding branch in `ios.py` makes exactly that test fail.

### Scope

**C9200L and C9300L.** Both families are documented fixed-uplink with no network-module slot, and both have a `show inventory` confirming the shape:

| family | why it qualifies | evidence |
|---|---|---|
| C9200L | every SKU has fixed uplinks ([data sheet](https://www.cisco.com/c/en/us/products/collateral/switches/catalyst-9200-series-switches/nb-06-cat9200-ser-data-sheet-cte-en.html)); the plain C9200 takes a removable `C9200-NM-*` | the reporter's own inventory in #556 |
| C9300L | no network-module slot, unlike the C9300 and C9300X ([data sheet](https://www.cisco.com/c/en/us/products/collateral/switches/catalyst-9300-series-switches/nb-06-cat9300-ser-data-sheet-cte-en.html)) | a published C9300L-24T-4X capture on IOS-XE 17.09.05 — four populated `Te1/1/x` rows, no FRU row — corroborated by an independent C9300L-48P-4X capture with the same structure |

Matching is on the PID prefix rather than link speed, so every uplink variant (`-4G`, `-4X`, `-2Y`) is covered without enumerating SKUs.

Further families should be added on the same footing: documentation saying the uplinks are fixed **and** a capture showing the inventory actually omits the FRU row. A wrong entry invents a chassis-level parent for hardware that already has one, which is the exact failure the surrounding gate exists to prevent.

### Fixture correction

Five `test_get_modules` fixtures gave a **C9300L chassis PID to devices that must have a removable uplink module** — hardware that cannot exist. Two of them also used `C9300L-NM-8X`, which is not a shipping part. Adding C9300L to the allowlist made that latent inaccuracy load-bearing, so the first commit corrects it:

| fixture | was | why it was wrong |
|---|---|---|
| `cat9300_stack_with_nm_uplinks` | C9300L + `C9300-NM-8X` | C9300L has no module slot |
| `fru_row_unusable_not_promoted` | C9300L + `C9300L-NM-8X` | same, and that PID does not ship |
| `mixed_fixed_and_uplink_transceivers` | C9300L + `C9300L-NM-8X` | same |
| `fru_row_omitted_not_promoted` | C9300L, no FRU row | pins a decline that is only correct when a module *could* have been there |
| `numeric_member_names_prefixed_mode` | C9300L, no FRU row | same |

All five move to the plain **Catalyst 9300**, the family that does take a `C9300-NM-*`. Behavior is unchanged at that commit — the same optics promote and the same optics are declined, now for reasons the hardware supports.

The last two are the ones that mattered. Left on a C9300L PID, adding the family to the allowlist flips them from decline to promote and **silently deletes the coverage they exist to provide**. Verified by reverting just those two fixtures with the new allowlist in place:

```
FAILED test_get_modules[fru_row_omitted_not_promoted]
FAILED test_get_modules[numeric_member_names_prefixed_mode]
FAILED test_get_modules_declines_promotion_when_fru_row_entirely_omitted
FAILED test_get_modules_numeric_member_names_enter_prefixed_mode
```

Splitting this into its own commit keeps that check reviewable: `a367c5f1` is green on its own, so the corrections are provably behavior-preserving before the allowlist entry lands on top.

### Tests

- `fixed_uplink_chassis_no_fru_row` — the C9200L reporter's inventory; both optics promote to device-rooted bays.
- `fixed_uplink_9300l_no_fru_row` — the sanitized C9300L-24T-4X capture; all four uplinks promote. It keeps the `StackAdapter1/x` rows a real capture carries, so the test also pins that they contribute no bays.
- Both carry `expected_result.json`, so the inherited `test_get_modules[scenario]` deep-compares the full payload.
- `test_optic_promotion_precedence.py` — the three precedence rules above, at helper level.

Dropping `C9300L` back out of the allowlist fails exactly the two new C9300L tests and nothing else, which is what makes the fixture corrections provably sufficient.

`2385 passed, 160 skipped`, ruff clean. Baseline before the change was `2378 passed`.

### Also not covered here

- **Blank-PID optics.** `Te1/1/3` and `Te1/1/4` on the reporter's device are DAC cables reporting a serial but no PID, skipped by design at `_parse_inventory_rows` because NetBox needs a model. They get 2 of 4 optics even with this fix.
- **Diagnosability.** When every optic is declined the whole path is DEBUG-only, so an operator who enables `discover_modules` and gets nothing has no signal at default log level. That is what turned this into a support ticket, and it deserves its own change.

The reporter's second device (WS-C2960S-24PD-L) is unaffected by this fix and by the allowlist: its uplinks sit on module 0 in both the `Te1/0/x` and `Gi1/0/25-26` representations, so it promotes under the baseboard rule introduced by #525.


leoparente:
@codex review

samiura (approved):

jajeffries (approved):

leoparente:
@codex review

PR #567, opened 2026-08-31
fix(device-discovery): stop netbox_id payloads sending placeholder site and role

leoparente:
Fixes [MON-331](https://linear.app/netboxlabs/issue/MON-331/device-discovery-undefined-siterole-placeholders-override-real-values). 

### What

`Defaults.site` and `Defaults.role` are pydantic defaults of the literal string `"undefined"`, and `translate_device` attached them unconditionally. Setting `netbox_id` on a scope only added the source_match metadata; it did nothing about the placeholders.

A payload carrying `source_match.netbox_id` binds an existing device by primary key, so the plugin skips matching and the defaults pass and applies **every field present in the payload**. The placeholder therefore repointed a real device's site and role at the `"undefined"` objects. `netbox_id` made the ENGHLP-751 overwrite deterministic instead of preventing it.

Measured end to end through `translate_data` with an empty policy:

| | site | role |
|---|---|---|
| before, with `netbox_id` | `'undefined'` | `'undefined'` |
| after, with `netbox_id` | **omitted** | **omitted** |
| after, no `netbox_id` | `'undefined'` | `'undefined'` |

### Why keyed on netbox_id rather than on the value being unset

The ticket offered two approaches and preferred option 1: change the defaults to `None` and omit the fields whenever they are unset. The code rules that out. **NetBox requires both `site` and `role` on a Device**, so a create has to carry something, and `netbox_id` is the only signal that separates an update from a create. Option 1 therefore collapses into option 2 for exactly these two fields.

That is not a judgment call, it is testable: suppressing unconditionally instead of only on `netbox_id` fails three pre-existing tests (`test_translate_device`, `test_translate_data`, `test_translate_data_handles_none_defaults_and_options`) plus the new guard. The constraint shows up directly.

This also matches `snmp-discovery` and `gnmi-discovery`, which omit site/role when the defaults are empty so netbox_id already gives a clean partial update there. device-discovery was the only backend with an always-attached placeholder.

### Also fixed

The placeholder rode along **inside `Location` and `Rack`**, both of which take a site. An explicit `location` with no configured site would have smuggled `site='undefined'` back into the payload through the nested object. Both now use the suppressed value, covered by its own test.

### Cleanup

`UNDEFINED_PLACEHOLDER` replaces the bare literal in the `Defaults` model so the sentinel has one authoritative home. `interface.py`'s prefix-scope cascade already compares against the same string and its comment already points at the Defaults model; I left that call site alone to keep the diff to the bug.

`_drop_unconfigured_placeholder` is a helper rather than inline branching because inlining pushed `translate_device` past the C901 complexity ceiling. It carries the reasoning above.

### Tests

Four new tests on the emitted proto, asserting presence rather than value since `site` and `role` are message fields:

- `netbox_id` + no configured site/role → both omitted (`HasField` false)
- `netbox_id` + explicitly configured site/role → both preserved, so the suppression cannot overreach
- no `netbox_id` → placeholders still sent, so creates keep working
- `netbox_id` + explicit location → the location does not carry the placeholder site

Both directions of the guard were mutation-verified: never suppressing fails the two suppression tests and nothing else; always suppressing fails the create-path tests.

`2402 passed, 160 skipped`, ruff clean. Baseline on develop was `2398 passed`.

### DIFF
--- a/orb-discovery/device-discovery/device_discovery/policy/models.py
+++ b/orb-discovery/device-discovery/device_discovery/policy/models.py
@@ -142,12 +142,21 @@ class PrefixParameters(IpamParameters):
     scope_location: str | None = Field(default=None, description="Prefix scope location, optional")
 
 
+#: Stand-in used when the operator configured no site or role. NetBox requires
+#: both on a Device, so a create has to carry something; everywhere else this
+#: string means "no value was configured" and is suppressed rather than sent.
+#: See translate_device and the prefix-scope cascade in interface.py.
+UNDEFINED_PLACEHOLDER = "undefined"
+
+
 class Defaults(BaseModel):
     """Model for default configuration."""
 
-    site: str | None = Field(default="undefined", description="Site name, optional")
+    site: str | None = Field(
+        default=UNDEFINED_PLACEHOLDER, description="Site name, optional"
+    )
     role: str | None = Field(
-        default="undefined", description="Device Role name, optional"
+        default=UNDEFINED_PLACEHOLDER, description="Device Role name, optional"
     )
     if_type: str | None = Field(default="other", description="Interface type, optional")
     interface_patterns: list[InterfacePattern] | None = Field(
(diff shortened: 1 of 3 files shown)


leoparente:
@codex review

samiura (approved):

PR #568, opened 2026-08-31
fix(device-discovery): budget target expansion across the policy, not per entry

leoparente:
Follow-up to #566. Reworked — the first version of this PR lowered the cap to 1024, which I no longer think was the right fix. See the comment above for why; in short, I had surveyed only one of the two siblings.

### What

#566 capped expansion at 65536 addresses **per scope entry**. That bounds one entry and nothing else:

```
16 x /16, each legal on its own  ->  policy expands to 1048544 addresses
```

So the unbounded expansion MON-328 was about is still reachable, just spelled differently. The cap needs to be a policy budget, not an entry ceiling.

### What the siblings do

| backend | limit | shape |
|---|---|---|
| gnmi-discovery | 1024 | per-policy, summed via `Count` before expanding |
| snmp-telemetry (#564) | 65536 | per-policy, `checkPolicyExpansion` |
| device-discovery (before) | 65536 | **per-entry** |

Both siblings budget the policy. device-discovery was the only one bounding a single entry, which is the actual divergence — not the number. `snmp-telemetry`'s comment states the case directly:

> The budget spans the policy rather than one entry. A per-target ceiling bounds a single target and nothing else: sixteen /16 entries each sit under it while the policy expands to 1048576 addresses.

**65536 rather than gnmi's 1024**, because the numbers differ for a reason. gnmi is 1024 because each address becomes a *persistent subscription*; snmp-telemetry is 65536 because each becomes a *recurring poll job*. device-discovery port-scans and then maybe opens an SSH session, which is the second cost model. Lowering to 1024 would have broken `/21` and `/20` scopes to fix nothing.

### How

`count_hostnames` reports what `expand_hostnames` would return without building any of it. It follows the same dispatch and calls the same endpoint parser, so no target shape can mean one thing to the budget and another to the expander. A test cross-checks the two across every accepted shape rather than asserting each in isolation, so a new branch in one has to be added to the other or the test fails.

The check is a model validator, so an oversized policy **fails the write** rather than failing later inside a scheduled job:

```
policy scopes expand to 1048544 addresses in total, more than the limit of 65536
```

The per-entry check stays as a backstop against the same constant, the way gnmi keeps `MaxExpand` inside `Expand`. One number, so there is one thing to reason about.

### Verified

```
16 x /16 (each legal alone)    REJECTED   1048544 addresses
one /24 + one /22              ACCEPTED
single /8                      REJECTED  16777214 addresses
```

Two mutations, each failing exactly the test that covers it and nothing else: summing changed to `max()` (the per-entry behaviour this PR replaces), and the validator removed entirely.

`2407 passed, 160 skipped`, ruff clean. Baseline on develop was `2398 passed`.

### Also

device-discovery had no documentation of expansion semantics at all, while gnmi documents them fully. Added: what a CIDR excludes and why a range does not, the IPv6 difference, the budget, and what a refused target does. Every number verified against the code.

### DIFF
--- a/docs/backends/device_discovery/README.md
+++ b/docs/backends/device_discovery/README.md
@@ -213,6 +213,28 @@ The scope defines a list of devices that can be accessed and pulled data.
 | override_defaults | map | no | Allows overriding of any defaults for a specific device in the scope |
 | netbox_id | integer | no | NetBox device primary key. When set, the diode plugin matches the device by PK instead of by name. Ignored when hostname is a subnet or IP range. |
 
+#### Subnet and range expansion
+
+A CIDR excludes its network and broadcast addresses, so `10.0.0.0/24` is 254
+addresses and `10.0.0.0/22` is 1022. A `/31` and a `/32` have no such pair to
+exclude and stay 2 and 1. IPv6 has no broadcast, so only the network address is
+excluded: `fd00::/126` is 3. A range excludes nothing, because someone who
+wrote `0` and `255` meant them, so `10.0.0.0-255` is 256.
+
+A policy may expand to at most **65536** addresses in total, counted across all
+its scope entries before any expansion happens. The budget spans the policy
+rather than one entry, so sixteen `/16` scopes are refused together even though
+each is under the limit on its own. A policy over the budget is rejected when it
+is written, naming the total and the limit:
+
+```
+policy scopes expand to 1048544 addresses in total, more than the limit of 65536
+```
+
+A single entry over the limit is also refused at expansion time as a backstop:
+it is skipped, named in an error, and recorded as a failed run, leaving the rest
+of the policy unaffected.
+
 ### SSH Configuration and Jumphost Support
 
 For advanced SSH scenarios including bastion/jumphost connectivity, VRF-aware connections, and multi-hop SSH configurations, see the dedicated guide: [SSH Configuration and Jumphost Support](./ssh.md).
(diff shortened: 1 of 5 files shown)


leoparente:
@codex review

leoparente:
Converting to draft — I based this on an incomplete survey and I think the change is wrong as written.

I compared device-discovery only against gnmi-discovery. There is a third answer in `orb-telemetry/snmp-telemetry` (PR #564, still draft so it is not on develop yet), and it is the most thoroughly reasoned of the three:

```go
const maxPolicyAddresses = 65536
```

with an argument that directly rebuts the shape of this PR:

> The budget spans the policy rather than one entry. A per-target ceiling bounds a single target and nothing else: sixteen /16 entries each sit under it while the policy expands to 1048576 addresses.

So the three backends line up as:

| backend | limit | shape |
|---|---|---|
| gnmi-discovery | 1024 | per-policy |
| snmp-telemetry | 65536 | per-policy |
| device-discovery | 65536 (#566) | **per-entry** |

Two things follow. The real divergence is the **shape**, not the number: device-discovery is the only one bounding a single entry rather than the policy, and that is exactly the hole the comment above describes. And the numbers plausibly differ for a reason — gnmi is 1024 because each address becomes a persistent subscription, snmp-telemetry is 65536 because each becomes a recurring poll job. device-discovery port-scans and then maybe opens an SSH session, which is the second cost model, not the first.

On that reading #566 had the right number and this PR is a regression, so lowering to 1024 fixes nothing real while breaking `/21` and `/20` scopes.

It also leaves MON-328 only half closed: a per-entry cap with unlimited entries is still unbounded in aggregate, which is the failure mode that ticket was about.

Reworking this into a per-policy budget of 65536, counted without expanding, with the per-entry check kept as a backstop the way gnmi keeps `MaxExpand` inside `Expand`.

leoparente:
@codex review

leoparente:
@codex review

samiura (approved):

PR #576, opened 2026-09-02
perf(snmp-telemetry): render the device tags on the first point that needs them

leoparente:
Closes MON-344.

`appendDeviceTags` walked every top-level `metric_tags` column on every cycle, consulting no poll window, and produced `baseAttrs` before any metric was collected. Those attributes are only ever consumed by points collected on that cycle, so a cycle that collects no point spent a credentialed SNMP request per tag column for nothing.

The attributes are now rendered on the first point that needs them and memoised for the rest of the run. A cycle that collects nothing walks nothing. The run polls one device from one goroutine, so the memo needs no synchronisation, which the comment says rather than leaving a reader to wonder.

## The ticket's impact estimate did not survive measurement, and the honest numbers are smaller

Worth stating plainly, because the ticket is what justified the work.

MON-344 claimed 65 of 202 profiles carry top-level tag columns and that a 10 second policy collecting hourly declarations wastes on the order of 360 tag walks per useful cycle. Measured against the resolved bundled set:

- **181 of 205** profiles carry top-level tag columns, not 65 of 202. The ticket's figure came from grepping `^metric_tags:` in source YAML, but `extends` appends `MetricTags` during resolution, so most profiles inherit them.
- **Essentially none can have a fully throttled cycle.** Exactly 1 of 205 declares `poll_time_sec` on every symbol; 11 declare no symbols at all; the remaining 193 always have at least one declaration due. So the throttling scenario the ticket described does not arise for the bundled set.

The saving is therefore not "cycles where everything is throttled". It is **cycles that collect no point at all**, which happens when a device answers none of the declarations that are due: a generic profile matched to a device implementing little of it, or a device that has stopped answering. That is a real and unremarkable situation, but it is narrower than the ticket said.

Measured on two profiles, driving two cycles with a walker that answers only sysObjectID, counting walks of each profile's own tag-column OIDs:

| profile | uncached tag columns | before | after |
|---|---|---|---|
| `aruba/aruba-clearpass.yml` | 12 | 12 per cycle | 0 |
| `arista/arista-switch.yml` | 3 | 3 per cycle | 0 |

"Before" was measured by applying the eager mutant, which is the behaviour on `develop`.

## The property this could have broken, and the tests that hold it

Commit `860946b9` fixed a defect where a device tag and a metric naming one OID were read from two different walks of it, so a point could carry an attribute from one moment beside a value from another. `minuteman-pdu.yml` produced a point showing one outlet both on and off. The run's walk cache is what fixed it.

Laziness inverts which side populates that cache: the metric now walks first and the tag reads the cached response, where before it was the other way round. It is one request and one response either way, so the property holds, but it is asserted rather than assumed. One test declares an OID as both a metric and a device tag, has it answer different values on successive walks, and asserts the OID is walked once, that the point carries the value and the attribute from the same response, and that the metric took the shared walk. A second does the same for a failed walk, since `860946b9` also made the two share a failure.

## A behaviour narrowing, stated rather than buried

A device tag walk can no longer be the sole cause of a run that collected nothing and failed a walk, because the tags are only walked once a point exists. One existing test asserted exactly that shape and failed outright; it now reaches the same guarded rule through a due declaration whose walk fails, and additionally asserts the tag column is not walked. Two other existing tests needed real changes for the same reason, including one that would have gone vacuous under laziness because its walker answered no metric OID.

## Verification

Every assertion added or edited, 37 of them, was flipped individually and re-run. Three named implementation mutants: rendering eagerly, which is today's behaviour, fails 6 tests; re-rendering per point rather than memoising fails the memo test, which is the only observable since the walk cache hides a repeated render from the wire; rendering before the poll-window check fails 4 on the scalar path and the table test on the other.

Existing device-tag tests were re-checked with a mutant that renders nothing: 19 fail, including all three walk-cache tests from `860946b9`. Two survive by design, asserting `reviewProfile` warnings rather than rendering.

Gates: build, vet, `-race`, `make lint` at 0 issues, and the root and `orb-discovery/snmp-discovery` modules still build.

### DIFF
--- a/orb-telemetry/snmp-telemetry/collector/collector.go
+++ b/orb-telemetry/snmp-telemetry/collector/collector.go
@@ -795,12 +795,16 @@ func (c *MetricsCollector) collect(ctx context.Context, key deviceKey, target co
 	walker.SetBulkWalk(!profile.NoUseBulkWalkAll)
 
 	// One walk per OID for the whole run, so declarations sharing a poll
-	// window share the walk that starts it. It is created before the device
-	// tags because they are read first and name OIDs the profile also declares
-	// as metrics, so the two have to be served by one response.
+	// window share the walk that starts it. A tag column can name an OID the
+	// profile also declares as a metric, and the cache is what serves the two
+	// of them from one response whichever asks first.
 	walks := make(walkCache)
 
-	baseAttrs := c.appendDeviceTags(ctx, walker, walks, profile, appendIdentityAttrs(nil, key), sysDescr, sysOIDValue)
+	// The attributes every series of this run carries, rendered when the first
+	// point needs them rather than here.
+	base := &deviceAttrs{render: func() []attribute.KeyValue {
+		return c.appendDeviceTags(ctx, walker, walks, profile, appendIdentityAttrs(nil, key), sysDescr, sysOIDValue)
+	}}
 
 	// fresh accumulates this run's observations, one per metric name and row.
 	// throttled records the declarations skipped because poll_time_sec has not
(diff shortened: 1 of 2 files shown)


leoparente:
@codex review

manrodrigues (approved):

PR #580, opened 2026-09-03
test(device-discovery): add the FortiOS 7.4.12 device capture

leoparente:
## Why

The 7.4.12 scenarios added with the #537 fix were reconstructed from the snippets in that issue, not captured from a device. Both their READMEs ended with "Replace this with a sanitised capture if the reporter supplies one."

The reporter has now supplied one, along with confirmation that the fix works on their FortiGate-100F.

## What the capture settles

The reconstructed fixtures had to guess at three things, and each README says so. The capture answers all three:

- **`switch:` and `aggregate:` positions and values.** Previously invented, or left out and covered only in unit tests. Now real, including `aggregate: LAN-LAG` on a member port and `switch:` on both a physical port and a VLAN.
- **`switch:` values can contain spaces.** One row carries a two-word value directly before `wccp: disable`, which is what actually exercises the field splitter. Nothing in the reconstructed data covered this.
- **`mtu-override:` really is last on 7.4.12.** The old README flagged this as unresolved, noting all eleven vendored ntc-templates captures put it before `wccp:`. The capture confirms the issue's ordering was not a transcription error.

Types covered by the flat capture: physical, tunnel, switch, aggregate, vlan, vxlan.

## Kept, not replaced

The reconstructed scenarios stay. The device puts `medium:` **below** `status:`, between `speed:` and `FEC:`. In that position, a regression that treats an unknown field line as unreadable and discards the block would still pass. The reconstructed fixture deliberately puts `medium:` above `status:`, where such a regression fails the test. The two scenarios test different things, so replacing one with the other would lose coverage. Both READMEs now explain this and cross-reference each other.

## Sanitising

- The reporter redacted eight addresses to `x.x.x.x`, which does not parse. Replaced with private and RFC 5737 documentation addresses, preserving the four-octet shape and the netmask beside it. `dmz` and `mgmt` were given the addresses the physical capture shows for the same interfaces.
- Two `switch:` values were site-specific names, replaced with neutral ones that keep the internal space.
- Nothing else altered: field order, indentation and trailing whitespace are as the device printed them. The physical capture needed no address changes at all.
- No `fnsysctl_ifconfig.txt` was supplied, and the mock device returns an empty string for a missing file, so MAC addresses come out empty rather than invented.

## Verification

`expected_result.json` for both scenarios was generated from the parser's actual output and checked by hand: 29 interfaces matching the 29 physical blocks, speeds and statuses correct, and only the 8 addressed interfaces in the IP result with `point1` correctly a /32.

```
89 passed, 4 skipped
```

Both new scenarios are picked up automatically by `parametrize_scenarios`, so no Python changed. The corpus parity fixtures are untouched: the capture cannot go there, because ntc-templates is exactly what fails to parse it.

### DIFF
--- a/orb-discovery/device-discovery/tests/custom_drivers/fortinet_fortios_ssh/mock_data/test_get_interfaces/7412/README.md
+++ b/orb-discovery/device-discovery/tests/custom_drivers/fortinet_fortios_ssh/mock_data/test_get_interfaces/7412/README.md
@@ -13,4 +13,7 @@ discards the block and this scenario fails.
 The issue's own field order is not evidence either way: its quoted flat line puts
 `mtu-override:` last, where all eleven vendored captures put it before `wccp:`.
 
-Replace this with a sanitised capture if the reporter supplies one.
+The reporter has since supplied a capture; it is the `7412_capture` scenario beside
+this one. This scenario is kept rather than replaced, because the device puts `medium:`
+below `status:`, where a regression that discards a block on an unknown field line would
+still pass. The deliberate ordering here is what catches it.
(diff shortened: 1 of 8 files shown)


leoparente:
@codex review

samiura (approved):

PR #601, opened 2026-09-10
fix(network-discovery): keep reverse hostnames within the characters NetBox accepts

leoparente:
## Summary

nmap prints an asterisk in place of any character it will not pass in a reverse-lookup name, a space inside a DNS label for instance, and lets a few others through (`+ = : ~`). The backend set that name as the IP's `dns_name` unchanged, and NetBox, which accepts only letters, digits, hyphens and underscores in a label and an asterisk only as a leading wildcard label, rejected the IP when the change was applied. Diode's server has no rule for the field, so nothing stopped the value on the way. ENGHLP-1668.

- **Every character NetBox rejects becomes a hyphen.** `abc1234-vendor*model*unit.example.net` arrives as `abc1234-vendor-model-unit.example.net`. nmap's asterisk never marks a wildcard, so it is not kept as one; a name NetBox already held with a leading asterisk changes on the next run.
- **A name with no acceptable form is left off** rather than guessed at: an empty label, more than 255 characters with the trailing dot, or nothing but hyphens after replacement. The address is still sent, and Diode's patch leaves a `dns_name` NetBox already holds.
- **The original is kept.** When a name was replaced or left off, and the comments are the backend's to write, the name as nmap gave it goes under the `hostnames` key of the IP's comments, a key that was always present and null. A name that needed no change records nothing, so existing IPs see no comment change. When the policy sets `defaults.comments` the backend does not add to them and logs the replacement at warning level instead.
- **Each run logs** how many names it replaced or left off.

The per-host entity build moved out of the run loop into `ipAddressEntity` so the wiring is testable.

Follow-up worth its own ticket: Diode's server validates nothing about `dns_name`, so every producer has to know NetBox's rule; a server-side rule would catch the next one.

## Testing

- `dnsName`: replacement in every label position, trailing dot kept, the 255-character boundary with and without the dot, refusals for empty labels and hyphen-only names. Fuzzed in review over nmap's output alphabet against NetBox's validator regex: no accepted output NetBox rejects.
- `applyHostname`: PTR preferred over the last name, only the used name recorded, nothing recorded when the policy owns the comments, an unusable name left off.
- `ipAddressEntity`: the recorded name lands in the comments JSON next to the ports; the policy's own comments go out untouched.
- Eleven killed mutants across the helper, the hook and the call site, including an inverted `canRecord` at the call site. Two adversarial review passes.
- `go test ./...` green apart from `TestRunStore_UpdateRun`, which compares two nanosecond timestamps and fails on develop the same way; `make lint`, 0 issues.
- No customer data in tests or docs: `example.net`, TEST-NET-1 addresses, invented names.

### DIFF
--- a/docs/backends/network_discovery.md
+++ b/docs/backends/network_discovery.md
@@ -8,6 +8,24 @@ The network discovery backend uses [Diode Go SDK](https://github.com/netboxlabs/
 
 IP addresses support VRF, Tenant, Role, Description, Comments and Tags via `defaults`.
 
+The name a reverse lookup returns for an address becomes the IP's `dns_name`,
+lowercased. NetBox accepts only letters, digits, hyphens and underscores in a
+label. nmap, which does the lookup, prints an asterisk in place of any
+character it will not pass, a space inside a DNS label for instance, and lets
+a few others through that NetBox rejects; every such character becomes a
+hyphen, so a name nmap printed as `abc1234-vendor*model*unit.example.net`
+arrives as `abc1234-vendor-model-unit.example.net`. A name with no form
+NetBox accepts, one with an empty label, longer than 255 characters or left
+with no letter or digit, is left off; the address is still sent, and a
+`dns_name` NetBox already holds for it stays as it is. In either case the
+name as nmap gave it is kept under `hostnames` in the IP's comments, unless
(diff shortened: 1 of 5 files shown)


leoparente:
@codex review

jajeffries (approved):

PR #602, opened 2026-09-10
feat(agent): wire gnmi-telemetry into the agent and ship it in the image

leoparente:
## Summary

The glue for the `gnmi-telemetry` backend, in the shape of #584 for `snmp-telemetry`. MON-356.

- **Backend package** `agent/backend/gnmitelemetry`: registry name `gnmi_telemetry`, exec `gnmi-telemetry`, API on `localhost:8079`. `common.otlp.grpc` is required and passed to `--otel-endpoint`; an agent that enables the backend without it refuses to start. `host`, `port`, `log_level`, `log_format`, `otel_export_period`, `policy_env_vars`, `profiles_root` and `profiles_dir` are each passed only when set with a non-empty value, an absent or null `host` or `port` keeps the default, non-string log keys are refused, the port is validated 1 to 65535. Stop hands the process the agent's default grace, which the binary sizes its whole shutdown to. No `PolicyStatusProvider`. The apply log carries only the policy's identity. The child's logfmt and JSON records are re-logged with their severity.
- **Registration** in `cmd/main.go`, with a test.
- **Image**: a `gnmi-telemetry` build stage stamping `GNMI_TELEMETRY_VERSION` and `BUILD_COMMIT`, copied into the final stage. No `USER` change.
- **Release parity**: `resolve-backend-versions` resolves `gnmi-telemetry/v*` tags; `release.yaml` detects changes under `orb-telemetry/gnmi-telemetry/`, waits on `gnmi-telemetry-release.yaml` and passes the version through; `_agent-release.yaml` gains the input, build arg and no-cache filter; `develop.yaml`, `build.yaml` and the Makefile pass the build arg. Until the first `gnmi-telemetry/v*` tag exists the image stamps `0.0.0`.
- **Docs**: `docs/backends/gnmi_telemetry.md`, the root README list and policies example, rows in `docs/configs/agent_yaml.md` including the env-var substitution table, and a section in `docs/config_samples.md`. Off by default: nothing is added to `default_config.yaml`.

Two things the adversarial review of this diff surfaced that stay for follow-ups: the package is a rename of `snmptelemetry` with the readers duplicated, which the ticket asked for file for file and which a shared set of plain reader functions in `agent/backend` should replace before a third telemetry backend; and the binary's own comments in `policy/sweep.go` and `server/server.go` say the agent's policy POST times out at 10 s where both telemetry backends use 30 s, a stale claim in the backend module left out of an agent-scoped PR.

## Testing

- Backend package tests: the args table, optional-flag reset, refused reconfiguration leaving the backend unchanged, IPv6 bracketing, stop grace, start and reset before configure refused, both log formats re-logged, an end-to-end apply, rename and remove against a recorded API, the server's rejection surfaced, and no policy secret in the log. Six killed mutants: OTLP endpoint defaulted instead of refused, `profiles_dir` under the snmp flag name, port above 65535 accepted, half the stop grace, policy body logged, default port 8078.
- Both documented policy samples parsed and validated by the binary's own `ParsePolicies` with the env var allowed, in a throwaway test run in the backend module.
- Under the race detector, with `make lint` at 0 issues.
- The `gnmi-telemetry` stage builds and `--help` runs. Full image: develop 940 MB, this branch 977 MB (the binary is 27 MB).
- End to end with the branch image on orb-test-lab: a `gnmi_telemetry` policy in a local agent config, credential through `policy_env_vars`, against the lab's gnmi-sim. The agent started the backend on 8079, applied the policy, and the collector received `gnmi.if_admin_status{device_ip, interface_name, netbox_id, policy}` plus `gnmi.target_up{mode=on_change}`, `gnmi.targets_active`, `gnmi.notifications_total` and `gnmi.updates_dropped_total`. The MON-354 containerlab topology is down, so the sim stood in.
- One adversarial review pass; its findings (a sample the binary rejected, three comments and a fixture inherited from the snmp package) are fixed in this revision.

### DIFF
--- a/.github/actions/resolve-backend-versions/action.yml
+++ b/.github/actions/resolve-backend-versions/action.yml
@@ -25,6 +25,9 @@ outputs:
   telemetry:
     description: 'Latest released snmp-telemetry version (X.Y.Z, or 0.0.0).'
     value: ${{ steps.resolve.outputs.telemetry }}
+  gnmi-telemetry:
+    description: 'Latest released gnmi-telemetry version (X.Y.Z, or 0.0.0).'
+    value: ${{ steps.resolve.outputs.gnmi-telemetry }}
 
 runs:
   using: composite
@@ -48,6 +51,7 @@ runs:
         dd=$(backend_version device-discovery)
         wk=$(backend_version worker)
         st=$(backend_version snmp-telemetry)
+        gt=$(backend_version gnmi-telemetry)
         {
           echo "network=${nd:-0.0.0}"
           echo "snmp=${sd:-0.0.0}"
@@ -55,4 +59,5 @@ runs:
           echo "device=${dd:-0.0.0}"
           echo "worker=${wk:-0.0.0}"
           echo "telemetry=${st:-0.0.0}"
+          echo "gnmi-telemetry=${gt:-0.0.0}"
         } >> "$GITHUB_OUTPUT"
--- a/.github/workflows/_agent-release.yaml
+++ b/.github/workflows/_agent-release.yaml
@@ -31,6 +31,10 @@ on:
         description: 'snmp-telemetry version to bundle (X.Y.Z).'
         required: true
         type: string
+      gnmi-telemetry-version:
+        description: 'gnmi-telemetry version to bundle (X.Y.Z).'
+        required: true
+        type: string
 
 permissions:
   contents: write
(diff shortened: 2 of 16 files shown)


leoparente:
@codex review

leoparente:
@codex review

jajeffries (approved):

PR #603, opened 2026-09-10
fix(agent): survive a fleet reset, a remove and a restart that reach a backend the agent did not start

leoparente:
## Summary

Every bundled backend is registered in `cmd/main.go`, but only the ones the agent's configuration names are configured and started, so on the image's default config five registered backends never receive a logger or a process. Three paths reached them and dereferenced that logger; all three are reachable from fleet RPCs on any agent running the shipped default config.

- **Fleet full reset.** The reset RPC walked the whole registry and called `FullReset` on every backend; pktvisor, opentelemetry_infinity and gnmi_discovery log before their process guard and panicked the agent. `backend.RestartAll` now leaves alone a backend that reports it was never started, which is the one meaning of `Unknown` (documented on the interface).
- **Policy remove.** `RemovePolicy` and the dataset removal behind it had no counterpart to the running check the apply path has, and called the backend with the nil logger; reachable from the remove action, the full-list reconciliation, a group removal and a dataset removal. Both now go through one helper that skips only a backend the agent never started, still removing the policy from the agent's repo as an expected condition logged at warning level. A backend that was started is asked whatever its state, since a live process whose status probe timed out may still hold the policy and the request's own timeout bounds the wait.
- **Restart wiping every policy.** `RemoveBackendPolicies` removed every policy of every backend from the repo when one backend restarted, so the others were gone until the next full list from fleet. It now takes the backend's name and removes only that backend's, through the same helper. `RestartBackend` also looks the backend up in the started set rather than the registry, so a restart requested for a registered but unstarted backend is refused instead of dereferencing a nil entry, and the restart counter tolerates a backend the monitor never registered.

Not in this PR, noted for a follow-up: after a restart the backend's own policies are not re-applied (`ApplyBackendPolicies` has no caller), so they come back only on fleet's next full list, on git's next scheduled refresh, or never under the local manager.

## Testing

- Six new tests, each paired with a killed mutant: a never-started backend skipped by the full reset, a restart counter for an unmonitored backend, a remove that does not call a never-started backend and logs no error, a remove that still asks a started backend whose status probe failed, a restart removing only its own backend's policies and going through the same gate, and a restart refused for a registered but unstarted backend.
- The three original crashes reproduced by an independent review against develop with the full nine-backend registry unconfigured, and confirmed gone on this branch by rerunning the same reproductions.
- Agent module suite green under the race detector; `make lint`, 0 issues.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -611,8 +611,11 @@ func (a *orbAgent) shutdownOTLP() {
 }
 
 func (a *orbAgent) RestartBackend(ctx context.Context, name string, reason string) error {
-	if !backend.HaveBackend(name) {
-		return errors.New("specified backend does not exist: " + name)
+	// Every bundled backend is registered; only the ones this agent started
+	// are in a.backends, and only those have a process to restart.
+	be, ok := a.backends[name]
+	if !ok {
+		return errors.New("backend is not started by this agent: " + name)
 	}
 
 	// Serialize concurrent Stop+Start sequences for the same backend across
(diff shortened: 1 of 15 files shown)


leoparente:
@codex review

leoparente:
@codex review

jajeffries (approved):

Issue #607, opened 2026-09-11
Orb always creates VLAN Group with Scope DCIM > Site


demianvdenb:
Hi there, I believe I just ran into a limitation of Orb.

## Context
The company where I work has multiple remote sites, and some sites have different names, but are in the same physical location.
So the sites in my NetBox instance are configured like this
Brussels -> mysite01
Brussels -> mysite02 
Antwerp -> mysite03
Where Brussels and Antwerp are Site Groups in NetBox.

I'm currently setting up NetBox to use it for IPAM as well, while already using it as an inventory of devices and cables.

## The problem
The Orb agent can detect VLANs that are defined on my switches using SNMP Discovery, but unfortunately VLANs with the same VLAN ID can have slightly different names across the switches.
This causes the Orb/Diode to create duplicate VLANs, e.g.
- ID:37 NAME: VLAN37
- ID:37 NAME: VLAN 37
- ID:37 NAME: vlan 37

This is where VLAN groups come in to play.
Within a VLAN group each VLAN ID should be unique, so my idea was to create a VLAN group for each Site Group with scope DCIM > Site Group.
I manually removed every duplicate VLAN and assigned each VLAN to a VLAN group.
Then in agent.yml I added this
```
snmp_discovery:
  switches_brussels:
    config:
      defaults:
        site: "Brussels"
        vlan:
          group: "Brussels VLAN Group"
```
When running this, Orb creates a new group with the name "Brussels VLAN Group" but with scope DCIM > Site, not DCIM > Site Group.
As far as I can see in the [Documentation](https://netboxlabs.com/docs/discovery/agent/backends/device_discovery/#nested-defaults), I can't make Orb create a VLAN Group with scope DCIM > Site Group.
And VLAN groups with DCIM > Site is not really an option for me, or I would have to completely overhaul my NetBox organization structure.

Thank you for having a look!

leoparente:
Hi @demianvdenb,

Thanks for the clear report. This was a limitation, not a configuration issue. Every discovery backend scoped the VLAN group to the site, and Diode matches a VLAN group on its name plus scope, so your site-group scoped group could never be matched and a second, site-scoped one was created instead.

`vlan.group` now also accepts a map with a scope. For your setup:

```yaml
snmp_discovery:
  switches_brussels:
    config:
      defaults:
        site: "mysite01"
        vlan:
          group:
            name: "Brussels VLAN Group"
            scope_site_group: "Brussels"
```

[...]

PR #610, opened 2026-09-11
feat(device-discovery): let vlan.group choose the NetBox scope of the VLAN group

leoparente:
## Summary

Device-discovery half of #607, same config shape as #609 for snmp-discovery.

- `vlan.group` keeps accepting a bare name, scoped to `defaults.site` as before.
- It also accepts a map with `name` and one of `scope_site`, `scope_site_group`, `scope_region` or `scope_location`. A location scope carries `defaults.site`, as prefix `scope_location` already does.
- A map with no scope behaves like the bare name. Two scopes at once, or a map without a name, fails policy validation with a clear message.
- A per-device `override_defaults.vlan.group` replaces the policy group as a whole. The generic deep merge would otherwise carry a policy `scope_site_group` into an override that only named a group.

## Config

```yaml
defaults:
  site: "mysite01"
  vlan:
    group:
      name: "Brussels VLAN Group"
      scope_site_group: "Brussels"
```

## Testing

- Model tests for the string form, the map form, two scopes and a missing name.
- Translate tests for each emitted scope type, explicit site over `defaults.site`, location carrying the site, location without a site, and the map-without-scope fallback.
- Override merge tests for replace-as-a-whole and for an override that leaves the group alone.
- `pytest` and `ruff check` pass on the module.

gnmi-discovery follows in a separate PR.

### DIFF
--- a/docs/backends/device_discovery/README.md
+++ b/docs/backends/device_discovery/README.md
@@ -233,13 +233,34 @@ Current supported defaults:
 | ├─ comments | str | VRF comments |
 | ├─ tags | list | VRF tags |
 | vlan       | map  | VLAN-specific defaults        |
-| ├─ group   | str  | VLAN group name. When set, every emitted VLAN is attached to an `ipam.vlangroup` scoped to `defaults.site`: the group's `scope_site` is populated from `defaults.site` |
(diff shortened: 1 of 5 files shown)


leoparente:
Applied the same unknown-key rule Codex raised on #609: the vlan.group map now rejects unknown keys (pydantic extra=forbid), so a misspelled scope fails validation instead of falling back to a site-scoped group. Commit 3deb826c.

leoparente:
@codex review

jajeffries (approved):

PR #611, opened 2026-09-11
feat(gnmi-discovery): let vlan.group choose the NetBox scope of the VLAN group

leoparente:
## Summary

gnmi-discovery part of #607, same config shape as #609 (snmp-discovery) and #610 (device-discovery).

- `vlan.group` keeps accepting a bare name, scoped to the device's site as before.
- It also accepts a map with `name` and one of `scope_site`, `scope_site_group`, `scope_region` or `scope_location`. A location scope carries the device's site, as NetBox locations are unique per site.
- A map with no scope behaves like the bare name. A per-target `override_defaults.vlan.group` replaces the policy group as a whole so a policy scope cannot leak in.
- Two scopes at once, a missing name, or an unknown key inside the map fails policy parsing. The unknown key case is stricter than the warn-only pass the rest of the policy gets, on purpose: that pass cannot see inside a custom decoder, and a typo like `scope_sitegroup` would otherwise silently create a site-scoped group that persists in NetBox.
- The gnmi docs had no row for `vlan` defaults at all; this adds it along with the group section.

## Config

```yaml
defaults:
  site: "mysite01"
  vlan:
    group:
      name: "Brussels VLAN Group"
      scope_site_group: "Brussels"
```

## Testing

- Config decoding tests for the string form, each scope, no scope, null, two scopes, missing name, unknown key, bad node kind and receiver reset.
- Merge test updated to check the override replaces the group as a whole.
- Builder tests for each emitted scope type, explicit site over the device site, location carrying the device site, location with no device site, and the map-without-scope fallback.
- `go test ./...` and `golangci-lint` pass on the module.

### DIFF
--- a/docs/backends/gnmi_discovery.md
+++ b/docs/backends/gnmi_discovery.md
@@ -71,8 +71,31 @@ gNMI discovery policies are broken into two subsections: `config` and `scope`.
 | interface | map | Interface defaults: `if_type` (fallback type, default `other`), `description`, `tags` |
 | ip_address | map | IP address defaults: `role`, `tenant`, `description`, `comments`, `tags` |
 | vrf | map | VRF defaults: `tenant`, `description`, `comments`, `tags` (name/RD come from discovery) |
+| vlan | map | VLAN defaults: `group` (see [VLAN group](#vlan-group)), `tenant`, `role`, `description`, `tags` |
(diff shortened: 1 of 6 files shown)


leoparente:
@codex review

jajeffries (approved):

Issue #606, opened 2026-09-11
EX4550-32F NETCONF — VLAN exists, but there are no port bindings.


ssgadzhi:
Title: Device discovery: Juniper EX4550 on Junos 15.1 emits VLANs but no interface VLAN associations

### Environment

- Device: Juniper EX4550-32F
- Junos: 15.1R7-S13
- Orb Agent: `2.14.0-09a4c81`
- Image: `netboxlabs/orb-agent@sha256:ae256574b3b3db5d0b657b1d4353200471a11c1913b70d77cb88ce043926d681`
- Backend: `device_discovery`
- Driver: `junos`
- Transport: NETCONF over SSH, port 830
- Test performed in dry-run mode; no data from this test was sent to Diode/NetBox.

### Problem

Discovery successfully retrieves device information and VLAN entities with the correct VLAN IDs. However, none of the emitted interface entities contains `mode`, `tagged_vlans`, or `untagged_vlan`.

The same symptom was observed with the earlier stable-image test and remains present in the development build above.

### Relevant policy settings

```yaml
config:
  defaults:
    site: "DRY-RUN"
    role: "Unclassified"
  options:
    capture_running_config: false
    capture_startup_config: false
    discover_vrfs: true
scope:
  - driver: junos
    hostname: "<device-management-IP>"
    username: "<discovery-user>"
    password: "${JUNOS_PASSWORD}"
    optional_args:
      port: 830
```

### Discovery results

```text
device:       1
interface:  139
ip_address:  51
prefix:      44
vlan:        39
vrf:         20
```

Device model, serial number and Junos version are retrieved correctly.

The VLAN output contains 38 numeric VLAN IDs matching `show vlans`, plus `default`, for which the projected `vid` is null.

### Device evidence

Relevant entries from `show vlans`:

```text
VL156    156
             xe-0/0/17.0*, xe-0/0/30.0*
VL178    178
             xe-0/0/17.0*, xe-0/0/30.0*
VL888    888
             ge-0/0/23.0*, xe-0/0/30.0*
```

The switch therefore has VLAN memberships that are absent from the discovery output.

### Output verification

```bash
jq '[
  .entities[]
  | select(.interface)
  | .interface
  | select(
      has("mode")
      or has("tagged_vlans")
      or has("untagged_vlan")
    )
]' <dry-run-output.json>
```

Result:

```json
[]
```

### Expected behavior

[...]

PR #618, opened 2026-09-11
fix(device-discovery): classify Junos switchports when the reply carries no port-mode element

leoparente:
address #606.

### The bug

A Juniper EX4550 on Junos 15.1 answers `<get-ethernet-switching-interface-information>` with all of its VLAN memberships and **no `<interface-port-mode>` element anywhere**. The parser derived access-versus-trunk from that element alone, so every port fell through to its "no VLAN config at all → routed" branch, discarding the memberships it had just read. The switch reached NetBox with 14 interfaces and not one `mode`, `tagged_vlans` or `untagged_vlan`.

Replaying the reporter's capture through the unmodified driver reproduces it exactly: **14 interfaces, 0 classified, 14 routed.**

This survived because every existing non-ELS fixture carries `interface-port-mode`, and none carries a unit suffix. The tests were built on synthetic XML that has neither property of the real reply, so the shape a real device produces was never exercised in either respect.

### The fix

**Ask for the reply that carries the mode.** The two CLI forms map to one RPC, the detailed one adding `<detail/>`. They do not answer alike: on this device the plain form has no mode element and the detailed form carries it on every interface. The detailed form is asked for first; the plain form still answers when it yields nothing, so a platform that rejects the argument keeps the behaviour it has, and an ELS switch that refuses both falls through to its own RPC as before.

**Read the mode off the membership where no element arrives.** A member the port carries tagged makes it a trunk; a member named `all` is a trunk-only construct (`vlan members all` is only configurable under `port-mode trunk`); a native VLAN id is only meaningful on a trunk, and decides the case a port with a single untagged member would otherwise get wrong. Otherwise an untagged member makes it access. This is the conclusion the SNMP path already draws from Q-BRIDGE.

Checked against the same device's detailed reply, which does carry the mode: **11 of its 11 classifiable interfaces agree, none disagree.** Inference stays the fallback rather than the rule, because the element states outright what membership can only imply.

**Resolve a member reported by name.** Junos reports some members by name with no tag id. The name is now put to the device's own VLAN table.

The match is **exact and never case-folded**, which matters more than it looks: on this switch the table holds `MGMT` at tag 20 while `me0.0` reports a member named `mgmt`, Junos having put the out-of-band port in a pseudo-VLAN of that name outside the switching space. Case-folding would bind the management port to VLAN 20, wrongly and silently. A name the table gives more than one id is refused for the same reason nothing else here guesses.

**Key the result by the physical port.** Junos reports switching per logical unit (`ge-0/0/23.0`), while NetBox carries switchport mode and VLANs on the port. `_physical_name` exists for exactly this and the ELS path has always normalised through it; this path did not. Since `apply_interface_vlans` pairs on an exact name and orb emits the port and the unit as separate Interfaces, every VLAN recovered here would have attached to the subinterface while the port stayed blank.

Member ids and the native VLAN id go through `coerce_vid`, so a value outside 1-4094 is refused as membership rather than counted and dropped later — a member with tagid 0 would otherwise make a port look like it had an untagged VLAN and write `mode=access` with no VLAN to go with it.

### Result on the reported device

| | interfaces | classified | routed |
|---|---|---|---|
| before | 14 | 0 | 14 |
| after, detailed reply | 14 | 14 | 0 |
| after, plain reply only | 14 | 11 | 3 |

The three that differ are the ones whose only member is nameless. With the detailed reply they arrive with the device's own `mode` and no VLAN; on the plain reply they stay routed. Worth stating plainly: that means `me0.0`, the out-of-band management port, is reported with `mode=access` and no untagged VLAN. That is the device's own statement rather than an inference, and the alternative is to contradict it.

Those three also still log a warning per discovery cycle. They are unresolvable by design, so the message is about data deliberately not emitted rather than a fault.

### Testing

The reporter's real reply shape is now a scenario, with VLAN names synthesized. Plus: the mode read off membership in each direction; `all`-only and native-VLAN-id membership; a VID outside the dot1q range from both a tag id and a resolved name; the detailed form being asked for first and the plain form still answering when it is refused; a name resolved from the table, one differing only in case, one the table gives two ids, and a table keyed by strings as PyEZ actually returns; and the table being fetched once at most and not at all when no member needs it.

A translation-level test crosses the seam neither side's own tests did: an RPC naming the unit, through the driver, to an assertion that the port carries the mode and VLAN and the subinterface beside it carries neither.

Ten mutations, each reverting one rule and failing only the tests that pin it.

The PyEZ test fake can now serve a distinct fixture for a boolean RPC argument, which is what lets one scenario hold both replies. Every existing scenario is unaffected: none has a flagged file, so the plain one still answers.

`uv run pytest tests/` and `uv run --with ruff==0.15.10 ruff check .` clean.

### DIFF
--- a/orb-discovery/device-discovery/custom_napalm/junos.py
+++ b/orb-discovery/device-discovery/custom_napalm/junos.py
@@ -200,7 +200,109 @@ def _els_details_to_switchports(root) -> dict[str, dict]:
     return result
 
 
-def _interface_to_switchport_info(intf_elem) -> SwitchportInfo:
+def _vlan_name_resolver(driver):
+    """
+    Return a callable mapping a VLAN name to its id, built on first use.
+
+    Junos reports a member either by tag id or by name alone. A name can only be
+    turned into an id by asking the device's own VLAN table, which costs an RPC,
+    so the table is fetched the first time a nameless member is actually seen and
+    not at all on the devices that never report one.
+
+    **Matched exactly, never case-folded.** On a measured EX4550 the VLAN table
+    holds ``MGMT`` at tag 20, while ``me0.0`` reports a member named ``mgmt``:
+    Junos puts the out-of-band management port in a pseudo-VLAN of that name that
+    is not in the switching space at all. Case-folding would bind the management
+    port to VLAN 20, which is wrong, and wrong silently. A name that differs in
+    case is a different name.
+
+    A name the table gives more than one id for is refused for the same reason
+    nothing else here guesses: the device is the only thing that could say which
+    was meant, and it has not.
+    """
+    cache: dict[str, int] = {}
+    loaded = False
+
+    def resolve(name: str) -> int | None:
+        nonlocal loaded
+        if not name:
+            return None
+        if not loaded:
+            loaded = True
+            cache.update(_vlan_ids_by_name(driver))
+        return cache.get(name)
+
+    return resolve
+
+
+def _vlan_ids_by_name(driver) -> dict[str, int]:
+    """Build name -> id from ``get_vlans()``, dropping names that are not unique."""
+    try:
+        vlans = driver.get_vlans() or {}
+    except Exception:
+        logger.debug("Junos get_vlans failed while resolving a VLAN name", exc_info=True)
+        return {}
+
+    ids_by_name: dict[str, set[int]] = {}
+    for vid, data in vlans.items():
+        name = (data or {}).get("name") or ""
+        parsed = _maybe_int(vid)
+        if not name or parsed is None:
+            continue
+        ids_by_name.setdefault(name, set()).add(parsed)
+
+    resolved: dict[str, int] = {}
+    for name, ids in ids_by_name.items():
+        if len(ids) > 1:
+            logger.warning(
+                "Junos VLAN name %r maps to %d ids (%s); not resolving members by that name",
+                name,
+                len(ids),
+                ", ".join(str(i) for i in sorted(ids)),
+            )
+            continue
+        resolved[name] = next(iter(ids))
+    return resolved
+
+
+def _members_to_vids(member_list, resolve_vlan_name) -> tuple[int | None, list[int], bool]:
+    """
+    Read one ``<interface-vlan-member-list>`` into (untagged, tagged, has-all).
+
+    A member carries its id in ``<interface-vlan-member-tagid>`` and its side of
+    the trunk in ``<interface-vlan-member-tagness>``. Where the id is absent the
+    name is put to the device's VLAN table; see _vlan_name_resolver.
+    """
+    members = _find_children(member_list, "interface-vlan-member") if member_list is not None else []
+
+    untagged_vid: int | None = None
+    tagged_vids: list[int] = []
+    has_all_member = False
+    for m in members:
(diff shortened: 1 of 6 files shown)


leoparente:
Valid, and fixed in 6cf34548.

You are right that `native_vid` was the one VID this PR left on `_maybe_int`, and the consequence is sharper than for the member ids because the previous commit gave an unusable native id two jobs at once: a trunk signal, and the value substituted for the untagged member. An interface with no mode element, a real untagged member, and a placeholder native id such as `0` was classified as a trunk, had `0` substituted for its member, and then had that rejected by `classify_switchport` — reporting a port that carries VLAN 888 as a trunk with no VLAN, and silently discarding the one it has.

It is also a regression rather than a pre-existing gap: before the inference existed that port simply read as routed, so nothing was written.

`native_vid` now goes through `coerce_vid`, matching the member ids. With the placeholder rejected, the port falls back to the untagged member and classifies as `access` on VLAN 888, which is correct.

Covered by `test_an_unusable_native_vlan_id_does_not_make_a_trunk`, mutation-verified: reverting the parse back to `_maybe_int` fails it and nothing else.

Worth noting the finding is well-aimed at this change's actual weak point. The gap sat one line above the member-id parse it was compared against, and my own comment explaining why range validation mattered there is what made the omission beside it conspicuous.

leoparente:
@codex review

leoparente:
Valid, and blocking was the right call. Fixed in b9034fed.

You are right on every point, including that `_physical_name` already existed for exactly this and that the ELS path has always normalised through it. This path keyed by the unit, so with `apply_interface_vlans` pairing on an exact name and orb emitting the port and the unit as separate Interfaces, every VLAN the PR recovered would have landed on `ge-0/0/23.0` while `ge-0/0/23` stayed blank. A fix that looks like it worked.

Two things worth recording about how it got this far.

It is the same root cause as the bug being fixed. Every synthetic non-ELS fixture was written without unit suffixes, so this path had never seen one; the real capture added in this PR is the first, and because it was a new expectation rather than an existing one, it pinned the wrong keys instead of failing. The fixtures lacked two properties of the real reply, not one.

And I steered the review away from it. I listed "the RPC keys match what `apply_interface_vlans` pairs against by exact name" as established ground truth. That is true of the logical unit and says nothing about where the configuration belongs, which is the actual question. Asserting it as settled is why neither review round questioned it.

What changed:

- `get_interfaces_vlans` keys by `_physical_name(ifname)`. The real capture now resolves to `ge-0/0/23`, `xe-0/0/17`, `ae4`, `me0`.
- The real-capture expectation is corrected to physical ports, as are the assertions in the new tests.
- Added the translation-level test you asked for, in `tests/test_translate.py`: an RPC naming `ge-0/0/23.0`, through the driver, asserting `ge-0/0/23` carries `mode=access` and `untagged_vlan=888` and that `ge-0/0/23.0` beside it carries neither. That seam is crossed by nothing else on either side.

Mutation-verified: reverting to the literal name fails the seam test on its own (`the driver must key on the physical port, got ['ge-0/0/23.0']`), and six other tests with it.

2474 tests pass, ruff clean. Noted on the human approval still being required.

leoparente:
@codex review

jajeffries (approved):

PR #621, opened 2026-09-12
feat(device-discovery): add device_name_source option to name devices by fqdn

### LINKED ISSUE #620: device-discovery: option to source Device.name from the fqdn fact
**Problem:**
translate_device() always names devices after the hostname fact from NAPALM's get_facts(); the fqdn fact is ignored. Organizations that name NetBox devices with FQDNs (a common DNS-based convention) currently have to embed the domain into the device host-name itself - e.g. on Junos set system host-name rtr1.dc1.example.net instead of the idiomatic short host-name + system domain-name pair - just so discovery produces the desired NetBox name.

**Proposal**
Add option *device_name_source* in device discovery policy, like:
```
config:
  options:
    device_name_source: fqdn   # default: hostname
```
With fqdn, device.name is taken from the fqdn fact, falling back to the hostname whenever the driver reported no usable FQDN:

- the fact is missing or empty;

- it is a known placeholder: "None" (the junos driver stringifies an undetermined fqdn) or "Unknown" (the ios-family default);

- it equals the hostname (every driver reports fqdn == hostname when no domain is configured), which adds nothing over the hostname.

The fallback matters: a naive fqdn or hostname would rename every ios-family device to Unknown.

**Notes/scope**

- Default stays hostname, so existing deployments are unaffected.
- Docs should call out that Diode matches devices by name: switching an existing deployment to fqdn creates new records unless the NetBox devices are renamed first (related: #167).
- Virtual-chassis member names built from *stack_member_name_template* keep using the raw hostname fact (member FQDNs generally don't exist); this is documented, and could be a follow-up if there is demand.
- snmp-discovery is out of scope: it has no fqdn fact (names come from sysName).

This slots in next to the existing name-shaping options (emit_device_name, stack_member_name_template).

misamoylov:
### Summary

Adds an opt-in policy option device_name_source: hostname | fqdn (default hostname) that selects which get_facts() fact populates Device.name. With fqdn, devices land in NetBox as `<host-name>.<domain>` while keeping the idiomatic short host-name + domain configuration on the device itself.

Fixes #620

### Usage

```yaml
policies:
  device_discovery:
    routers:
      config:
        defaults:
          site: DC1
        options:
          device_name_source: fqdn
      scope:
        - driver: junos
          hostname: 192.0.2.1
          username: admin
          password: ${PASS}
```

### Behavior

`_resolve_device_name()` uses the `fqdn` fact only when it is genuinely usable, and falls back to the hostname otherwise:

| fqdn fact value | resulting Device.name |
|---|---|
| `rtr1.dc1.example.net` | `rtr1.dc1.example.net` |
| missing / empty | hostname |
| `"None"` (junos stringifies an undetermined fqdn) | hostname |
| `"Unknown"` (ios-family default placeholder) | hostname |
| equal to hostname (no domain configured) | hostname |

The blank-omission rule is preserved: no usable hostname *and* no usable fqdn then `Device.name` is omitted, not sent empty (same contract as before).

Because the resolved name feeds `device_params` before the interface/IP deepcopy, every nested device stub inherits it — no separate stub handling needed. `emit_device_name: false` still suppresses the name afterwards, orthogonally.

### Out of scope (documented)

- Virtual-chassis member names from `stack_member_name_template` keep using the raw hostname fact (member FQDNs generally don't exist). Could be a follow-up if there is demand.
- snmp-discovery: no fqdn fact there (`sysName`-based naming).

### Backward compatibility

Default is `hostname` - behavior is unchanged unless the option is set. Docs note that Diode matches devices by name, so switching an existing deployment to `fqdn` creates new records unless the NetBox devices are renamed first.

### Testing

- New `tests/test_device_name_source.py`: default/None options keep hostname; fqdn source; every fallback case; blank-name omission; invalid option value rejected by pydantic.
- Full backend suite: `pytest tests/` - 2470 passed, 160 skipped, 6 warnings.
- `ruff check` clean on the touched files. (`black` left untouched lines alone; baseline files are not black-clean under current black, so no unrelated reformatting was included.)

### Changes

- `device_discovery/policy/models.py` - `Options.device_name_source: Literal["hostname", "fqdn"]`, default `hostname`.
- `device_discovery/translate.py` - `_resolve_device_name()` helper + use it for `device_params["name"]`.
- `tests/test_device_name_source.py` - new.
- `docs/backends/device_discovery/README.md` - options table row.


leoparente:
Hi @misamoylov,

Thanks for this, and for the write-up in #620. The motivation is clear, the fallback logic is well thought through, and the tests cover it. We want to take it in. Two things need to change first, and I have a few smaller suggestions.

**Two issues to address**

1. The placeholder list cannot be complete. Our Palo Alto driver reports `fqdn` as `N/A`, and NAPALM's IOS driver reports `<hostname>.not set` when no domain is configured, because `show hosts` prints "Default domain is not set" and the driver keeps the text after "is". Both pass the current guards, so a Palo Alto policy would name every firewall `N/A` and, since name is the primary matcher, collapse them into one NetBox record. A positive check is safer than a denylist: use the fqdn only when it contains a dot, has no whitespace, and starts with the hostname followed by a dot.

2. Virtual chassis renames its master on every stack. The member copy in `translate_chassis._build_member_devices` keeps the raw `fqdn` fact and only rewrites `hostname` with the rendered template, so the equality guard compares against `core-sw-1` and never fires. Without a domain the master becomes `core-sw` while members stay `core-sw-2`; with a domain it becomes the FQDN. Either way `unique_master` resolves a different device and a new chassis is created. Resolving the name once from the raw facts, before the stack split, avoids it. Our preference is that stacks keep the template for every member including the master, with a doc line saying FQDN naming does not apply to stacks. A chassis test would have caught this; `tests/test_translate_chassis.py` already pins `core-sw-1` for the master.

**Suggestions**

- Guard the double-domain case. IOS, Junos, Check Point and Aruba OS build `hostname.domain` with no check, so an operator who already set `rtr1.dc1.example.net` as the hostname and also has a domain gets `rtr1.dc1.example.net.dc1.example.net`. Keeping the hostname when it already contains a dot is cheap, and the docs should mention it.
- Compare hostname and fqdn case-insensitively, and normalise the option value with a warning the way `emit_prefix_vlan` does rather than rejecting the whole policy on `FQDN`.
- Add tests for the chassis path, for nested device stubs on interfaces and IPs carrying the fqdn, and for `emit_device_name: false` combined with `fqdn`. A test that runs `_resolve_device_name` over every driver's `test_get_facts` expected result would prove no driver yields a bad name.
- The `emit_device_name` description in models.py and the README still say the name comes from the hostname fact.

On naming: we did consider folding this into `emit_device_name` to keep the option count down, and decided against it. That option mirrors snmp-discovery's, which has no fqdn fact, and it carries a matcher safety guard that should stay a plain on/off switch. Keeping `device_name_source` as you have it is the right call.

misamoylov:
> Hi @misamoylov,
> 
> Thanks for this, and for the write-up in #620. The motivation is clear, the fallback logic is well thought through, and the tests cover it. We want to take it in. Two things need to change first, and I have a few smaller suggestions.
> 
> **Two issues to address**
> 
>     1. The placeholder list cannot be complete. Our Palo Alto driver reports `fqdn` as `N/A`, and NAPALM's IOS driver reports `<hostname>.not set` when no domain is configured, because `show hosts` prints "Default domain is not set" and the driver keeps the text after "is". Both pass the current guards, so a Palo Alto policy would name every firewall `N/A` and, since name is the primary matcher, collapse them into one NetBox record. A positive check is safer than a denylist: use the fqdn only when it contains a dot, has no whitespace, and starts with the hostname followed by a dot.
> 
>     2. Virtual chassis renames its master on every stack. The member copy in `translate_chassis._build_member_devices` keeps the raw `fqdn` fact and only rewrites `hostname` with the rendered template, so the equality guard compares against `core-sw-1` and never fires. Without a domain the master becomes `core-sw` while members stay `core-sw-2`; with a domain it becomes the FQDN. Either way `unique_master` resolves a different device and a new chassis is created. Resolving the name once from the raw facts, before the stack split, avoids it. Our preference is that stacks keep the template for every member including the master, with a doc line saying FQDN naming does not apply to stacks. A chassis test would have caught this; `tests/test_translate_chassis.py` already pins `core-sw-1` for the master.
> 
> 
> **Suggestions**
> 
>     * Guard the double-domain case. IOS, Junos, Check Point and Aruba OS build `hostname.domain` with no check, so an operator who already set `rtr1.dc1.example.net` as the hostname and also has a domain gets `rtr1.dc1.example.net.dc1.example.net`. Keeping the hostname when it already contains a dot is cheap, and the docs should mention it.
> 
>     * Compare hostname and fqdn case-insensitively, and normalise the option value with a warning the way `emit_prefix_vlan` does rather than rejecting the whole policy on `FQDN`.
> 
>     * Add tests for the chassis path, for nested device stubs on interfaces and IPs carrying the fqdn, and for `emit_device_name: false` combined with `fqdn`. A test that runs `_resolve_device_name` over every driver's `test_get_facts` expected result would prove no driver yields a bad name.
> 
>     * The `emit_device_name` description in models.py and the README still say the name comes from the hostname fact.
> 
> 
> On naming: we did consider folding this into `emit_device_name` to keep the option count down, and decided against it. That option mirrors snmp-discovery's, which has no fqdn fact, and it carries a matcher safety guard that should stay a plain on/off switch. Keeping `device_name_source` as you have it is the right call.

Thanks for the thorough review. All addressed in the latest commit with new tests.

leoparente (approved):

PR #626, opened 2026-09-14
feat(snmp-discovery): read HUAWEI-VLAN-MIB VLAN catalog on Huawei devices without Q-BRIDGE-MIB

samiura:
## Why

Huawei access platforms such as the SmartAX MA5600T/MA5608T OLTs implement no
Q-BRIDGE-MIB at all: `dot1qVlanStaticTable`, `dot1qPortVlanTable` and
`dot1dBasePortIfIndex` all answer "No Such Object". Discovery completed cleanly
but emitted zero VLAN entities, even though the device publishes its VLAN
database under HUAWEI-VLAN-MIB `hwVlanMIBTable` (`.1.3.6.1.4.1.2011.5.6.1.1`).

## What

- New `huawei` vendor gate on sysObjectID prefix `.1.3.6.1.4.1.2011.`, so the
  table is walked only on Huawei hosts (same pattern as the Cisco and Juniper
  gates).
- Walk `hwVlanIndex`, `hwVlanName` and `hwVlanRowStatus` behind it. The index
  column establishes which VLANs exist — a VLAN with no configured description
  has no `hwVlanName` row at all. Names join the same fallback tier as the Cisco
  VTP catalog, with `dot1qVlanStaticName` still authoritative where a device
  answers both. `hwVlanRowStatus` is a standard RowStatus and feeds the existing
  active/reserved derivation. Index-only rows are gated by `create_unknown_vlans`
  exactly like status-only dot1q rows.
- Docs: supported-platforms table and the snmp_discovery README.

## Scope

Only the catalog is read: VLAN IDs, names and status. Interface
`mode` / `tagged_vlans` / `untagged_vlan` remain unresolved on these devices —
`hwVlanPorts` is a bridge-port PortList and the OLT publishes nothing to
translate it with, and the mapper deliberately refuses a bridge-port == ifIndex
guess. The existing "missing dot1dBasePortIfIndex; skipping interface mutations"
warning now fires on these OLTs, since VLAN rows are present.

## Testing

- Vendor-matcher cases for SmartAX and VRP sysObjectIDs, plus a near-miss arc.
- mapping.yaml structural test asserting the three columns land in the huawei
  walk set and in no other.
- VlanMapper tests built from a real MA5608T walk: 13 VLANs emitted, 5 with
  device names and the rest as `VLAN<vid>`; `create_unknown_vlans: false` keeps
  only the named 5; RowStatus → status; dot1q precedence over the vendor name;
  no interface mutation through the post-pass.
- `go test -race ./...` and `golangci-lint` clean for the module.


samiura:
@codex review

leoparente:
Should we isolate the HUAWEI-VLAN-MIB handling into vlan_huawei.go, similar to the existing vendor-specific VLAN logic? I think VlanMapper should remain the common orchestrator/emitter, but having Huawei OIDs and catalog parsing spread through vlanNamesByVid, mergeVLANNames, emitVLANs, and hasVLANSignal makes the generic mapper increasingly vendor-aware. A Huawei extractor returning normalized VID/name/RowStatus data would keep the precedence/emission logic generic and give us a clean place to extend Huawei port membership later.

leoparente:
**P2: make VLAN RowStatus precedence deterministic when both Q-BRIDGE and HUAWEI-VLAN-MIB answer**

For VLAN names we explicitly make dot1qVlanStaticName authoritative over the Huawei catalog, but RowStatus does not have the same precedence.

Right now both dot1qVlanStaticRowStatus and hwVlanRowStatus eventually do:

ensure(vid).rowStatus = st

while iterating over ObjectIDValueMap. Since this is a Go map, if both MIBs expose a status for the same VID, whichever happens to be visited last wins.

That means a VLAN could theoretically alternate between e.g. active and reserved across runs purely due to map iteration order.

Can we make this deterministic, ideally with the same rule as names — standard Q-BRIDGE data wins when both are present — and add a test with conflicting Huawei/dot1q RowStatus values?

samiura:
> Should we isolate the HUAWEI-VLAN-MIB handling into vlan_huawei.go, similar to the existing vendor-specific VLAN logic? I think VlanMapper should remain the common orchestrator/emitter, but having Huawei OIDs and catalog parsing spread through vlanNamesByVid, mergeVLANNames, emitVLANs, and hasVLANSignal makes the generic mapper increasingly vendor-aware. A Huawei extractor returning normalized VID/name/RowStatus data would keep the precedence/emission logic generic and give us a clean place to extend Huawei port membership later.

Done — Huawei handling now lives in vlan_huawei.go behind a vendor-neutral vlanCatalog (VIDs / names / RowStatus) that the generic mapper consumes without knowing the MIB. vlan_mapper.go has no Huawei-specific prefixes left, and there's a natural place for port membership once we have a translation source for hwVlanPorts.

> **P2: make VLAN RowStatus precedence deterministic when both Q-BRIDGE and HUAWEI-VLAN-MIB answer**
> 
> For VLAN names we explicitly make dot1qVlanStaticName authoritative over the Huawei catalog, but RowStatus does not have the same precedence.
> 
> Right now both dot1qVlanStaticRowStatus and hwVlanRowStatus eventually do:
> 
> ensure(vid).rowStatus = st
> 
> while iterating over ObjectIDValueMap. Since this is a Go map, if both MIBs expose a status for the same VID, whichever happens to be visited last wins.
> 
> That means a VLAN could theoretically alternate between e.g. active and reserved across runs purely due to map iteration order.
> 
> Can we make this deterministic, ideally with the same rule as names — standard Q-BRIDGE data wins when both are present — and add a test with conflicting Huawei/dot1q RowStatus values?

Good catch. Status now follows the same rule as names: the vendor catalog is applied first and dot1qVlanStaticRowStatus overwrites it, so precedence comes from pass ordering rather than map iteration. Added a repeated-run test that fails on the previous code.

leoparente (approved):

Issue #629, opened 2026-09-15
Juniper EX4550-32F: SNMP discovers VRFs but emits IP addresses without VRF associations


ssgadzhi:
A full dry-run comparison on our EX4550-32F running Junos 15.1R7-S13 shows missing IP-to-VRF associations in SNMP discovery.

Environment:

Orb Agent: 2.14.0-07af109

SNMP Discovery: 1.36.1

Protocol: SNMPv2c

Both comparison runs used dry-run mode

Results:

SNMP: 17 VRF entities, 47 IP addresses, none of the IP addresses has a VRF reference.

NETCONF: 20 VRF entities, 51 IP addresses, 37 have VRF references, including 33 IPv4 addresses.

All 47 SNMP address/prefix-length strings are also present in the NETCONF output. The four additional NETCONF addresses are IPv6.

For example, the management address on lo0.24 belongs to MGMT-TRSP in NETCONF, but the same address/interface has no VRF reference in SNMP. The SNMP device's primary IP also lacks that reference.

The SNMP INFO log reports successful crawl completion and ingestion of 239 entities, with no timeout or VRF-specific error. It contains separate warnings about an unknown model OID and out-of-range MTU values; we have not established any connection between those warnings and this issue.

Expected: IP addresses belonging to discovered VRFs should retain their VRF associations.

Could you investigate the membership mapping and specify any additional SNMP tables needed for diagnosis?

The VLAN fix from #605 remains confirmed. This is a separate IP/VRF association issue.

leoparente:
Hi @ssgadzhi,

Thank you, the NETCONF control makes this unambiguous and saves a lot of guessing.

VRF names and VRF membership come from two different tables. Names and route distinguishers come from `mplsL3VpnVrfTable`, and interface membership from `mplsL3VpnIfConfTable`, which is indexed by VRF name plus ifIndex. The IP association is driven entirely by the second one: an address inherits the VRF of the interface it is assigned to, and with no membership rows there is nothing to inherit. So 17 VRF entities with zero references on the addresses is exactly the shape of a device that answers the name table and not the membership table.

We checked our side before asking you for anything. Replaying a recorded Junos walk that publishes both tables gives 2 VRFs and 7 interface memberships, parsed correctly, so the reader handles the Junos index encoding. What we cannot tell from here is whether your EX4550 publishes that table at all. Junos support for this MIB differs between the routing platforms and the EX switches, and 15.1 is old enough that it may simply not be there.

[...]

PR #632, opened 2026-09-15
feat(snmp-discovery): read VRF interface membership from JUNIPER-VPN-MIB

leoparente:
address #629

A Junos EX4550 on 15.1 discovers its VRFs over SNMP and gives none of its IP
addresses a reference to them. The reporter's NETCONF run against the same
device returns 37 of those associations, so the gap is this backend rather than
the device.

## Cause

VRF names and VRF membership come from two different tables, and the IP
association is driven entirely by the second: an address inherits the VRF of the
interface it is assigned to. The reporter walked all three tiers:

```
mplsL3VpnVrfTable    1.3.6.1.2.1.10.166.11.1.2.2   populated, all 17 VRFs
mplsL3VpnIfConfTable 1.3.6.1.2.1.10.166.11.1.2.1   No Such Object
pre-standard arc     1.3.6.1.3.118.1.2.1           No Such Object
```

So every tier produced names and none produced membership, and `AttachVrfs` had
nothing to attach.

## What changed

`jnxVpnIfTable` carries it. Its index is `{ vpnType, vpnName, ifIndex }`, which
is the standard tier's `{ vrfName, ifIndex }` with a type in front, so the
existing `decodeOctetStringIndexWithTail` reads it once the type is removed. It
becomes a fourth tier behind the two standard arcs and the Cisco one, tried only
where none of them produced membership, and gated to Juniper.

Only `bgpIpVpn` rows. The same table carries `l2Circuit`, `bgpL2Vpn` and
`bgpVpls`, which are not VRFs: on the reported device its three `l2Circuit` rows
are named after interfaces, and its `jnxVpnTable` holds 30 VPNs against the
standard table's 17. Taking them would attach an L2 circuit's interfaces to
whatever VRF happened to share its name.

Names are carried as well as membership, so a device publishing this table and
no standard one still yields VRFs. Where the standard table has already named
them, `mergeVrfRecords` adds the membership and adopts nothing new, so this
table's L2 entries stay out of the emitted catalog without a second guard.

Nothing below this changes: `AttachVrfs`, the primary-IP resync and the prefix
derivation all already work, and were simply never handed a membership map on
this device.

## Evidence

No walk in the 1877-file corpus publishes this table, on any Juniper platform,
so there was nothing to build against and nothing that could regress. The
reporter walked their own device's `jnxVpnIfTable` and reported its shape, which
is what the tests are built to:

- 31 rows, 28 of them `bgpIpVpn` across 13 VRFs, 3 `l2Circuit`
- `l2Circuit` names are interface-shaped on that device, not VRF-shaped
- VPN name lengths run from 4 to 21 bytes, which is what exercises the
  length-prefixed index
- the reporter confirmed those 13 names match `mplsL3VpnVrfTable` byte for byte,
  including case and punctuation
- the remaining 4 of the 17 have no membership rows at all, and still emit as
  VRFs carrying no interfaces

The rows are written out in the test rather than committed as a captured file,
which is how every other test in this package builds its OID maps.

## Testing

- `collectJuniperVrfs` over the reported shape: 13 VRFs, 28 memberships,
  `l2Circuit` rows excluded
- end to end on the reported shape, including a VRF the standard table names
  with no membership anywhere
- the tier ordering, with a Juniper row for a VRF the standard tier already
  placed on an interface it did not, which must be ignored
- every rule mutation-verified: reverting one fails its own tests and no others
- `go test ./...` and `golangci-lint run` clean; 0 recorded devices change

The reporter has offered to test the resulting develop image against the same
EX4550 and compare it with their NETCONF control.


leoparente:
@codex review

samiura (approved):

PR #636, opened 2026-09-17
Fix: Link to lookup files in readme

Joshinken:
link to vendor files previously pointed to an invalid location


leoparente (approved):

leoparente:
hi @Joshinken thanks for the contrib

PR #642, opened 2026-09-17
feat(snmp-discovery): read the device serial from vendor chassis scalars when ENTITY-MIB has none

samiura:
## Summary

The device serial comes from `entPhysicalSerialNum` on the ENTITY-MIB chassis row. Two shapes leave that with nothing usable, and both emitted the device with `serial: null` even though the device publishes it through an enterprise scalar:

- a platform that implements no `entPhysicalTable` at all (pre-ELS Junos switches such as the EX4550);
- a platform whose chassis row carries an empty serial while serial-column strings appear only on contained components such as USB controllers (RouterOS).

This walks `JUNIPER-MIB::jnxBoxSerialNo` and `MIKROTIK-MIB::mtxrSerialNumber` under their vendor gates and, on the path where chassis inventory produced no usable row, takes the device serial from whichever answered.

## Behaviour

- **ENTITY-MIB keeps priority.** The fallback only runs when no chassis row with a serial exists, so platforms that publish the standard column (QFX5100, for example) and Virtual Chassis master pinning are unaffected.
- **Component serials are never selected.** Strings on contained rows (`tilegx-ehci.0`, `rb400_usb`) were never chassis-class rows at the root and remain excluded.
- **Vendor-gated walks.** MikroTik joins the vendor matcher table (`.1.3.6.1.4.1.14988.`); each scalar is only asked of its own vendor's targets. Non-Juniper/MikroTik targets see no change in walk set.
- Nothing is invented: when neither source answers, `serial` stays unset.

## Testing

- Unit tests for both device shapes, standard-column priority, a two-member VC ignoring the scalar, empty/NUL values, walk-set gating, `mapping.yaml` declarations, and the full mapper → chassis-translation pipeline.
- Dry-run against snmpsim replaying real RouterOS, QFX5100 and QFX5130 recordings: develop emits `serial: null` on all three; this branch emits the vendor serial on each, with entity counts unchanged.
- Docs: new "Device serial" section in supported_platforms.md and a note in the stack section of the README.


samiura:
@codex review

leoparente (approved):

PR #649, opened 2026-09-18
docs(agent,gnmi-discovery): drop beta and experimental labels, and the develop-image requirement

leoparente:
The CyberArk, Delinea Secret Server and DSV pages carried a "Beta" notice and
told readers to pull `netboxlabs/orb-agent:develop` "until a release containing
this integration is published". All three have shipped, so the published docs
were telling users to run an unreleased image and describing released features
as provisional. These pages are the source for
`netboxlabs.com/docs/discovery/agent/secretsmgr/...`.

## Beta notices

They mixed two different kinds of statement.

The capability limits are facts a user needs, and they stay, as plain notes:

- Delinea: read-only, username and password authentication only
- DSV: read-only, client-credential authentication only

The rest was internal QA status, naming which fakes the integration was tested
against and saying end-to-end validation was still to come. That belongs in the
repository, not on a page someone reads while configuring a secrets manager.
The CyberArk notice contained only that, so it is removed rather than reworded.

## Develop image

`docker pull netboxlabs/orb-agent:develop` becomes `:latest` in the Delinea and
DSV walkthroughs, and the prerequisite line that explained the temporary
substitution is now just "Docker".

## Validation walkthroughs

Kept, since verifying a secrets manager against a real tenant is genuinely
useful, but retitled away from test-plan language: "Manual end-to-end
validation" to "Verifying the integration", and "Pass criteria" to "Expected
results".

## gnmi-discovery experimental marker

`gnmi-discovery` was marked experimental in two places, the orb-discovery
README and the releasable-backend list in CONTRIBUTING. It cuts its own release
and ships in the aggregated agent release alongside every other backend, none
of which carry the marker. Its own README has no such wording, so those two
were the whole of it.

## Which Delinea deployments each option fits

Raised by a customer whose Secret Server tenant is on `delinea.app` rather than
`secretservercloud.com`. The docs described `server_url` as the on-prem and
Platform option and `tenant` as the cloud one, which reads as a
hosted-versus-self-hosted choice. It is not.

`tenant` is a shorthand that expands to a fixed hostname,
`https://<tenant>.secretservercloud.com`, so it fits that one domain and
nothing else. `server_url` is used as given and works for every deployment,
including a cloud tenant on another domain.

A cloud customer therefore read the docs and picked the one setting that cannot
reach them. Delinea is moving tenants to `delinea.app`, and regional tenants on
`secretservercloud.eu` and similar have always been in the same position, so
this is a growing set rather than an edge case. Both options now state which
hostnames they cover, and the rule is given directly under the sample config.

No behaviour change: `server_url` already supports these deployments.

## Left alone deliberately

- `docs/backends/snmp_discovery/README.md` calls MPLS-VPN-MIB "the pre-standard
  experimental arc". That is the MIB's own name, since it lives under the IETF
  experimental tree, not a claim about our maturity.
- `docs/backends/device_discovery/supported_platforms.md` says `aruba_aoscx_ssh`
  is "not yet supported for VRF discovery". That is a real capability gap, not
  provisional framing.

Docs only, no code or behaviour change.


samiura:
lint failure, I think it is the title

samiura (approved):

PR #653, opened 2026-09-21
fix(snmp-telemetry): let a sysobjectid wildcard match the arc it names

leoparente:
## Cause

A profile's `sysobjectid` wildcard `X.*` was indexed as the prefix `X.` and matched with `HasPrefix`, so it accepted strict descendants of `X` only. ktranslate, whose profiles these are, resolves a device by probing `<sysObjectID>.*` first and then successively shorter prefixes, so upstream `X.*` covers `X` itself.

A device that reports the very arc a vendor wildcard names therefore fell through to the generic `1.3.6.1.4.*` profile and lost the vendor's metrics. Every RouterOS device does this: its sysObjectID is `.1.3.6.1.4.1.14988.1` and the MikroTik profile is written as `1.3.6.1.4.1.14988.1.*`. The lab's RB5009 simulator polled with `generic/base.yml`, 14 IF-MIB metrics, no CPU, memory or PoE.

## Change

A wildcard now covers the arc it names as well as the arcs below it (`oid == X || HasPrefix(oid, X+".")`). The dot after the prefix still keeps a sibling arc that shares leading digits out, so `1.3.6.1.4.1.14988.10` does not match `1.3.6.1.4.1.14988.1.*`. Exact entries still win, and among wildcards the longest prefix still wins, so an arc that is both a narrow wildcard's own arc and inside a broad one goes to the narrow one.

## Effect on the bundled profiles

The change can only add matches for an OID equal to some wildcard's arc; exact entries and longer wildcards keep precedence, so nothing that matched before matches less. Of the 159 wildcard arcs in the bundled set, 134 resolve differently when a device reports exactly that arc:

| from | to | arcs |
|---|---|---|
| `generic/base.yml` | the vendor profile that owns the arc | 70 |
| no profile | `_template.yml` for its placeholder arc, and `generic/base.yml` for the `1.3.6.1.4` root it claims itself | 2 |
| `juniper/juniper-all-devices.yml` | the Juniper model profile whose variant arc it is (SRX 23, EX 15, MX 14, QFX 1); all four extend `juniper-all-devices.yml`, so the device keeps every series and gains the model's | 53 |
| a vendor's broader profile | the narrower profile of the same vendor (`apc_ups` to `apc_pdu` for `318.1.3.4`, `netgear-generic` to `readynas` and `netgear-access-point`, `Ruckus_Contoller_SNMP` to `ruckus-wap` and `ruckus-unleashed`, `meraki-cloud-controller` to `meraki-device`) | 9 |

Each destination is the narrowest wildcard whose arc the device reports, which is what ktranslate resolves for the same profiles.

Across the 1,754 LibreNMS recordings that carry a sysObjectID, 26 change verdict, all from `generic/base.yml`: every RouterOS recording (13) to `mikrotik-router.yml`, every SwOS recording (4) to `mikrotik-switch.yml`, four APC management cards that report the bare APC arc to `apc_ups.yml` (the profile that claims the enterprise arc), two Tripp Lite to `tripplite-ups.yml`, two ServersCheck to `infrasensing/sensor-gateway-base-unit.yml`, one Nutanix to `nutanix.yml`.

## Verification

Unit: a wildcard matches its own arc in both the device spelling and the profile spelling, a sibling arc sharing leading digits and the arc above stay out, the own arc keeps longest-prefix precedence and loses to an exact entry, and on the bundled set the MikroTik and APC arcs reach their vendor profiles from the generic one, three arcs move from a vendor's broader profile to its narrower one, and the neighbours that must not move do not. Each assertion is paired with a killed mutant (strict descendants only; no dot guard; trailing-dot prefix). One test that inspected the index's internal prefix form was updated to the new form with the same intent. Three adversarial review rounds; the README's matching paragraph now states that a wildcard covers its own arc.

Lab, snmp-telemetry built from this branch against the five snmpsim devices, develop vs branch:

| device | develop | this branch |
|---|---|---|
| RouterOS RB5009 (`.1.3.6.1.4.1.14988.1`) | generic/base.yml, 14 metrics | mikrotik-router.yml, 20 metrics (adds CPU, memory, PoE status/power/voltage) |
| Cisco 2960X, Cisco IOS, Junos SRX100, Junos QFX5100 | unchanged | unchanged, same profile and metric counts |


leoparente:
@codex review

jajeffries (approved):
Nice!

PR #654, opened 2026-09-22
fix(gnmi-telemetry): open one stream per origin, since a target takes one origin per Subscribe

leoparente:
## Cause

A target takes one origin per Subscribe RPC. The backend put every subscription of a profile into one `SubscribeRequest`, each path with its own origin, and the SR Linux profile deliberately mixes the OpenConfig base paths with one native-origin path (`/platform/control[slot=*]/memory`, `origin: ""`). SR Linux refuses that request outright, `Unimplemented: Mixing of different origins is not supported for Subscribe RPC`, on both streaming rungs, so the ladder fell to Get polling. Get skips the native-origin subscription by design, and SR Linux answers an OpenConfig wildcard list path with the whole subtree nested under an ancestor path, which the flat-leaf model cannot read; only `/system/memory/state`, answered at its own path, came through. The profile written for SR Linux is what disabled streaming on SR Linux: interface counters, CPU and native memory never reached the collector.

## Change

A target's subscriptions are grouped by origin and each group streams on an RPC of its own on the same session. The attempt stays one for the ladder and the consumer:

- its channels carry every stream's notifications;
- one sync response goes out once every stream has answered its own, naming every path they carry between them, so the consumer reconciles once against the whole dump as before;
- the first stream to fail or end ends the attempt, the others torn down with it, so a rung is refused or reconnected for the whole profile and never half-served.

The session keeps one handle per stream; `StopSubscribe` and `Close` release every handle and end the attempt, which is what releases a forwarder still waiting to deliver to a caller that has stopped reading. A single-origin attempt takes the previous path unchanged.

Two things the consumer could see on one stream and cannot see across several, the session now says:

- an error a stream reports before serving any data of its own is marked `gnmi.ErrBeforeData`, the target's code kept, and the consumer reads a mode rejection so marked as a refusal however much another stream delivered, so the ladder walks instead of reconnecting on the refused rung for ever;
- each stream keeps the initial-dump deadline the consumer keeps for a single stream, since another stream's data would keep the consumer's armed and the per-stream syncs never reach it; a stream that answers neither data nor its sync in time ends the attempt as `gnmi.ErrStreamSilent` when it served nothing, which the ladder advances through, or as a plain stall when it did, which the same rung reconnects through;
- a stream the target ends after accepting it ends the attempt with a plain error rather than a bare close, since the consumer judges a bare close from its own view of the attempt, which never saw the sync that stream answered; the same rung reconnects through it.

## Verification

Unit, against the package's in-process gNMI server taught to refuse mixed origins the way SR Linux does: two origins open two requests of one origin each and deliver one sync naming both paths, after the slower stream synced; one origin still opens one request; a refused origin ends the attempt with its error, the other stream torn down and the channels closed; `StopSubscribe` closes every stream on the target; `StopSubscribe` and `Close` each release forwarders waiting on a caller that stopped reading, observed by the attempt's goroutines going away; a stream refusing after another served is marked as before its data with the code kept; a stream that never syncs ends the attempt within the probe deadline, silent or stalled by whether it served; a stream the target ends after its sync, in either sync order, ends the attempt with a plain error and never a bare close. In the collector, a marked rejection and a silent stream after another stream's data each walk the ladder to the next rung. Each behaviour is paired with a killed mutant. Both packages under the race detector. Four adversarial review rounds; each of the first three found one real gap in how the split attempt reads to the consumer, all fixed above.

Lab, gnmi-telemetry built from develop and from this branch against the containerlab spike topology (two real Nokia SR Linux nodes) and the canned gnmi-sim, one `mode: auto` policy, `origin: openconfig`:

| target | develop | this branch |
|---|---|---|
| SR Linux, two nodes | `get` after both stream rungs refused; 3 metrics (`memory_physical`, `memory_reserved`, `target_up`) | `on_change`; 30 metrics, 383 points per export: interface counters, oper and admin status for 59 interfaces, CPU per core, OpenConfig and native memory, temperature, component status |
| gnmi-sim (single origin) | `on_change`, `arista_eos`, `if_admin_status` + `target_up` | unchanged |

Not changed: the Get rung still cannot read a subtree answered under an ancestor path, which now only matters on a target that refuses every streaming mode.


leoparente:
On the AI review finding (a sibling stream ending the attempt mid-handoff can drop the notification in flight, so the consumer's `productive` may not reflect a stream that served, and its error would be read as a refusal): applied in a845172e. The dropped notification itself is harmless, since the attempt is ending and the reconnect dumps again, but the verdict must not turn on it. The session now marks an error from a stream that had served data as `gnmi.ErrAfterData`, the twin of `ErrBeforeData`, and the consumer reads it as plain whatever it received; with the two marks and `ErrStreamSilent`, no error the fan-in delivers is classified from the consumer's own view. Tests: the after-data mark on the session side, and in the collector a marked rejection with nothing received keeps its rung; both assertions fail against the previous code.

leoparente:
@codex review

jajeffries (approved):

PR #655, opened 2026-09-22
docs(agent): bring the pktvisor reference into the repo

leoparente:
## Summary

The pktvisor backend page linked out to `orb.community` for everything beyond the agent-side settings: the policy schema, the inputs, the handlers and the metrics. That site is retired, so the reference those links pointed at is no longer reachable.

This brings that material into `docs/backends/pktvisor/` as plain markdown, one page per handler plus shared pages for inputs, handlers and metrics, and adds a netprobe page the old site never had.

It is not a copy. The upstream pages were written for the Orb control plane and for a single 3,500-line MkDocs page, and a lot of what they said is not true of the binary this agent ships. Every claim here was checked against `netboxlabs/pktvisor` at the revision the agent runs (`PKTVISOR_TAG ?= develop-alpine`), not against the old docs.

## What the old docs got wrong

These would each have cost a reader time, and all are corrected here:

- **Window settings were documented under `handlers.config`.** That key is never read. `HandlerManager` reads only `modules` and `window_config`, so every documented "abstract configuration" was a silent no-op.
- **`kind` was marked optional.** `pktvisord` rejects a policy without it, and the agent does not inject one.
- **A tap has no `filter` key.** `TapManager` reads only `config` and `tags`, so a filter written on a tap is dropped *without an error* and the policy captures everything.
- **The policy envelope was the control plane's** (`name`, `backend`, `schema_version`, `format`). The agent supplies the wrapper itself, so a policy is just `input`, `handlers`, `config` and `kind` under `orb.policies.pktvisor.<name>`.
- **`require_version` defaults to `1.0`**, so `type: dns` alone runs the 1.0 handler. The old page led with 2.0 and never said this.
- **Config keys that do not exist**: flow's `first_filter_if_as_label`, and `exclude_ips_from_summarization_flow` (that suffix only existed in the old page's HTML anchors). Both are rejected by `validate_configs`.
- **`targets` was documented on the netprobe handler.** It belongs to the input, is required there, and the handler rejects it.
- **Twelve fabricated metric names** in the Network 1.0 table, and `dns_wire_packets_dot`/`_doh`, which are incremented but never serialized.
- **`recorded_stream: false` enables it** — all eight handlers check presence and never read the value.
- **A quantile's `_sum` series carries the maximum, not a sum.**
- **Response code 2 is `SRVFAIL`, not `SERVFAIL`** — and `only_rcode` looks names up in that very table.

## New: netprobe

The netprobe input accepts 28 stream config keys and 5 per-target keys; the old page covered 7 of them and two of its five test types. [`input_netprobe.md`](docs/backends/pktvisor/input_netprobe.md) documents ICMP, TCP and HTTP in full, with DNS over HTTPS alongside them since it shares the transport settings.

Three things worth knowing from it:

- `test_type: udp` passes config validation, then fails at start with `Test type currently not supported` — no probe is built for it.
- HTTP follows redirects **only** when neither per-target `headers` nor a request `body` is set, so setting either silently turns following off and reports the 30x as the result.
- Scope enforcement is not uniform: the HTTP checks, `body`, `proxy` and `tls` are rejected off their test type, but the packet-pacing keys are accepted anywhere and simply ignored off `ping`.

## Also

Documents `input_resources`, a handler no policy names — `pktvisord` attaches it to every input stream on its own — which upstream never documented at all.

Drops the install and "running the agent" pages, which this repo already covers, and the Orb cloud agent configuration section.

## Upstream bug found on the way

The 2.0 DNS handler implements `only_xact_directions` but omits it from its `_config_defs`, so `validate_configs` rejects any policy that sets it. The filter is documented with that caveat. Worth filing against pktvisor separately.

## Validation

13 files, 3,411 lines. Every internal link and anchor resolves under GitHub/MkDocs slug rules, all 141 YAML examples parse, no duplicate heading slugs, no ragged tables. The repo README and `docs/configs/agent_yaml.md` point at the new path.

Reviewed adversarially in four full rounds plus three scoped verification passes; roughly 55 defects found and fixed, the last pass clean.


leoparente:
@codex review

samiura (approved):

jajeffries (approved):

PR #661, opened 2026-09-23
feat(snmp-telemetry): export one instrumentation scope per policy

leoparente:
## Summary
snmp-telemetry now exports every series of a policy under its own OpenTelemetry instrumentation scope, `snmp-telemetry` with the scope attribute `policy_name=<policy>`, the same convention pktvisor uses. The `policy` attribute is gone from every datapoint; datapoints carry only device dimensions (`device_ip`, `device_port`, `netbox_id`, `snmp_context`, profile tags). The platform side reads one convention across backends and the agent hop no longer repeats the policy on every point.

This is the first of three PRs: snmp-telemetry (this one), gnmi-telemetry (same change), then the agent's OTLP bridge stamping `orb.policy_id` / `orb.backend` on each `policy_name` scope. No compatibility work: the binary has no external users yet and ships inside the agent image at the same version.

## What changed
- `metrics`: `PolicyMeter(policy)` returns the process scope with `policy_name` as an instrumentation attribute; the SDK exports one `ScopeMetrics` per policy. `GetMeter()` stays for process-level metrics. The test hook is now provider-based (`SetMeterProviderForTest`) so tests read per-policy scopes through a manual reader.
- `collector`: the shared `MetricsCollector` keys its observable gauges by (policy, metric name) and registers each on the policy's meter; a new `policyDevices` index under `storeMu` lets each callback walk only its policy's devices. `ForgetPolicy` gives the policy's callbacks back to the meter (previously they lingered until `Close`). `appendIdentityAttrs` no longer appends `policy`; `policy` stays a reserved profile tag name so no profile can put the label back on the datapoint.
- `traps`: `snmp.traps_received{device_ip, trap_name, version}` is one counter per policy on the policy's meter, attached on `Activate` (lease granted) and detached on `Withdraw` (lease released); `Register` also attaches counters for policies activated before it ran. `snmp.traps_dropped` and `snmp.traps_datagrams` stay on the plain scope. `Close` no longer holds the registration lock across `Unregister`.
- Docs: README and `docs/backends/snmp_telemetry.md` describe the per-policy scope, the datapoint attributes, and that the SDK keeps a meter per distinct policy name for the process lifetime.

## How tested
- `GOWORK=off go test -race ./...` green across the module; `GOWORK=off make lint` 0 issues; `make test-coverage` total 90.9%.
- New manual-reader tests: two policies export in two scopes with no `policy` datapoint attribute; the same device under two policies stays two series; `ForgetPolicy` silences one scope, leaves the other, and a same-name policy registers afresh; `policyDevices` stays in step with `deviceStore` through collect, forget-device, forget-policy and close; traps attach on `Activate`, detach on `Withdraw`, attach on a late `Register`, process counters stay unscoped; `Unregister` is proven to run with no collector lock held, and a race test runs `Activate`/`Received`/`Withdraw` against a collecting reader.
- Adversarial review passes per commit plus a whole-branch review before opening; the last commits are the review fixes.

## Follow-ups (not in this PR)
- The trap tally's per-policy callback still scans all of `t.received` (bounded by `maxSeries`, about 10k entries) per active policy per export; indexing it by policy like the collector is a cheap follow-up if trap volume grows.
- `Acquire`/`Release` in the trap pool hold `pool.mu` while registering or unregistering a callback, which can wait for one in-flight collection. No lock cycle; latency only.


leoparente:
@codex review

samiura (approved):

PR #662, opened 2026-09-23
feat(gnmi-telemetry): export one instrumentation scope per policy

leoparente:
## Summary
gnmi-telemetry now exports every series of a policy under its own OpenTelemetry instrumentation scope, `gnmi-telemetry` with the scope attribute `policy_name=<policy>`, the same convention pktvisor uses and snmp-telemetry adopts in #661. The `policy` attribute is gone from every datapoint, including `gnmi.target_up`; datapoints carry only device dimensions (`device_ip`, `netbox_id`, promoted path keys). The platform reads one convention across backends and the agent hop no longer repeats the policy on every point.

Second of three PRs: snmp-telemetry (#661), gnmi-telemetry (this one), then the agent's OTLP bridge stamping `orb.policy_id` / `orb.backend` on each `policy_name` scope. No compatibility work: the binary has no external users yet and ships inside the agent image at the same version.

## What changed
- `metrics`: `PolicyMeter(policy)` returns the process scope with `policy_name` as an instrumentation attribute; the SDK exports one `ScopeMetrics` per policy. `GetMeter()` stays for the process-level health counters. Provider-based test hook (`SetMeterProviderForTest`).
- `collector/store`: series are keyed by (metric, policy, attributes) instead of reading the policy back from a `policy` attribute, so two policies on one device with identical attributes are two series. Delete, reconcile (evict) and forEach take the policy explicitly; a delete for one policy's target can no longer withdraw another policy's series on the same device.
- `collector/export`: observable instruments are keyed by (policy, metric name) and registered on the policy's meter; each callback reads only that policy's series. `forgetPolicy` gives a policy's callbacks back to the meter (previously they lingered until close). The schema registry and the process-wide series budget stay keyed by metric name.
- `collector`: `gnmi.target_up` is one gauge per policy on the policy's meter, with `device_ip` and `mode` only; `ForgetPolicy` unregisters it with the policy's series callbacks, outside every collector lock. `baseAttrs` no longer carries `policy`; `policy` stays a reserved promoted-attribute name so no profile can put the label back.
- Docs: README and `docs/backends/gnmi_telemetry.md` describe the per-policy scope, the datapoint attributes, the new `target_up` attribute set, and that the SDK keeps a meter per distinct policy name for the process lifetime.

## How tested
- `GOWORK=off go test -race ./...` green across the module; `GOWORK=off make lint` 0 issues; `make test-coverage` total 85.0%.
- New manual-reader tests: two policies export in two scopes with no `policy` datapoint attribute; the same device under two policies stays two series, in the store and in the export; delete and evict are scoped to the policy; `ForgetPolicy` silences one scope (series and `target_up`), leaves the other, and a same-name policy registers afresh; `target_up` is one point per policy under its own scope; `Unregister` is proven to run with no collector lock held.
- Adversarial review passes per commit plus a whole-branch review before opening; the last commits are the review fixes.


samiura (approved):

leoparente:
@codex review

leoparente:
@codex review

leoparente:
On the AI review's "per-policy-name meter cache grows without bound" finding (metrics.go `PolicyMeter`):

Acknowledged, and it is by design rather than an oversight: the `PolicyMeter` doc comment and the README both state that the SDK keeps a meter, plus one instrument per exported metric name, for every distinct policy name seen over the process lifetime, and that a restart clears it. Two points on the size of the risk:

- The SDK's `MeterProvider` has no API to evict a meter, so `ForgetPolicy` cannot release it; what it does release is everything that costs memory at scale. After the callbacks are unregistered the observable instruments' aggregators are emptied each cycle and nothing is exported for that scope (the `ForgetPolicy` tests assert the scope is gone from the export), so a stale name retains a meter object and empty instrument shells, on the order of kilobytes per name, not per series.
- Policy names are stable, user-chosen identifiers in all three strategies (YAML mapping keys, Git file names, fleet policy names), and the same name is reused across re-applies and restarts. An automation that mints a fresh name per apply would churn every per-policy structure in the agent (repo entries, backend policies, trap leases), not only this cache; we do not support that pattern.

The same trade-off applies to snmp-telemetry (#661), documented the same way. If we want a signal for name churn, a process-level gauge of distinct policy names seen is a small follow-up; happy to open a ticket if you think it is worth it.

